PR #22, opened 2021-09-04
Match Monero 0.17.2.3 decoy selection algo

j-berman:
Matches https://github.com/monero-project/monero/pull/7845 && https://github.com/monero-project/monero/pull/7821

### DIFF
--- a/src/util/gamma_picker.cpp
+++ b/src/util/gamma_picker.cpp
@@ -40,6 +40,8 @@ namespace lws
     constexpr const double gamma_shape = 19.28;
     constexpr const double gamma_scale = 1 / double(1.61);
     constexpr const std::size_t blocks_in_a_year = (86400 * 365) / DIFFICULTY_TARGET_V2;
+    constexpr const std::size_t defaut_unlock_time = CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE * DIFFICULTY_TARGET_V2;
+    constexpr const std::size_t recent_spend_window = 50 * DIFFICULTY_TARGET_V2;
   }
 
   gamma_picker::gamma_picker(std::vector<uint64_t> rct_offsets)
@@ -49,7 +51,7 @@ namespace lws
   gamma_picker::gamma_picker(std::vector<std::uint64_t> offsets_in, double shape, double scale)
     : rct_offsets(std::move(offsets_in)),
       gamma(shape, scale),
-      outputs_per_second(0)
+      seconds_per_output(0)
   {
     if (!rct_offsets.empty())
     {
@@ -59,8 +61,11 @@ namespace lws
       const std::size_t outputs_to_consider = rct_offsets.back() - initial;
 
       static_assert(0 < DIFFICULTY_TARGET_V2, "block target time cannot be zero");
+      // match wallet2's integer truncation implementation to avoid creating distinct anonymity puddles
       // this assumes constant target over the whole rct range
-      outputs_per_second = outputs_to_consider / double(DIFFICULTY_TARGET_V2 * blocks_to_consider);
+      seconds_per_output = DIFFICULTY_TARGET_V2 * blocks_to_consider / outputs_to_consider;
+      if (seconds_per_output == 0)
+        seconds_per_output = double(DIFFICULTY_TARGET_V2 * blocks_to_consider) / outputs_to_consider;
     }
   }
 
@@ -84,17 +89,27 @@ namespace lws
     static_assert(std::is_empty<crypto::random_device>(), "random_device is no longer cheap to construct");
     static constexpr const crypto::random_device engine{};
     const auto end = offsets().end() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
+    const uint64_t num_rct_outputs = spendable_upper_bound();
 
     for (unsigned tries = 0; tries < 100; ++tries)
     {
-      std::uint64_t output_index = std::exp(gamma(engine)) * outputs_per_second;
-      if (offsets().back() <= output_index)
+      double output_age_in_seconds = std::exp(gamma(engine));
+
+      // match wallet2 implementation to avoid creating distinct anonymity pools
+      // shift output back by unlock time to apply distribution from chain tip
+      if (output_age_in_seconds > defaut_unlock_time)
+        output_age_in_seconds -= defaut_unlock_time;
+      else
+        output_age_in_seconds = crypto::rand_idx(recent_spend_window);
+
+      std::uint64_t output_index = output_age_in_seconds / seconds_per_output;
+      if (num_rct_outputs <= output_index)
         continue; // gamma selected older than blockchain height (rare)
 
-      output_index = offsets().back() - 1 - output_index;
+      output_index = num_rct_outputs - 1 - output_index;
       const auto selection = std::lower_bound(offsets().begin(), end, output_index);
       if (selection == end)
-        continue; // gamma selected within locked/non-spendable range (rare)
+        throw std::runtime_error{"Unable to select random output - output not found"};
 
       const std::uint64_t first_rct = offsets().begin() == selection ? 0 : *(selection - 1);
       const std::uint64_t n_rct = *selection - first_rct;
--- a/src/util/gamma_picker.cpp
+++ b/src/util/gamma_picker.cpp
@@ -41,7 +41,7 @@ namespace lws
     constexpr const double gamma_scale = 1 / double(1.61);
     constexpr const std::size_t blocks_in_a_year = (86400 * 365) / DIFFICULTY_TARGET_V2;
     constexpr const std::size_t default_unlock_time = CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE * DIFFICULTY_TARGET_V2;
-    constexpr const std::size_t recent_spend_window = 50 * DIFFICULTY_TARGET_V2;
+    constexpr const std::size_t recent_spend_window = 15 * DIFFICULTY_TARGET_V2;
   }
 
   gamma_picker::gamma_picker(std::vector<uint64_t> rct_offsets)
--- a/src/util/gamma_picker.cpp
+++ b/src/util/gamma_picker.cpp
@@ -40,7 +40,7 @@ namespace lws
     constexpr const double gamma_shape = 19.28;
     constexpr const double gamma_scale = 1 / double(1.61);
     constexpr const std::size_t blocks_in_a_year = (86400 * 365) / DIFFICULTY_TARGET_V2;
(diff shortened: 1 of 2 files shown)


src/util/gamma_picker.cpp
```diff
@@ -84,17 +89,27 @@ namespace lws
...
     for (unsigned tries = 0; tries < 100; ++tries)
     {
-      std::uint64_t output_index = std::exp(gamma(engine)) * outputs_per_second;
-      if (offsets().back() <= output_index)
+      double output_age_in_seconds = std::exp(gamma(engine));
+
+      // match wallet2 implementation to avoid creating distinct anonymity pools
+      // shift output back by unlock time to apply distribution from chain tip
+      if (output_age_in_seconds > defaut_unlock_time)
+        output_age_in_seconds -= defaut_unlock_time;
```
vtnerd, line 101: Looks like I need to go back over the review that just occurred in `wallet2`.

This is creating a bias in the selection algorithm. It is now more likely to select blocks 10-20.

src/util/gamma_picker.cpp
```diff
@@ -59,8 +61,11 @@ namespace lws
       const std::size_t outputs_to_consider = rct_offsets.back() - initial;
 
       static_assert(0 < DIFFICULTY_TARGET_V2, "block target time cannot be zero");
+      // match wallet2's integer truncation implementation to avoid creating distinct anonymity puddles
       // this assumes constant target over the whole rct range
-      outputs_per_second = outputs_to_consider / double(DIFFICULTY_TARGET_V2 * blocks_to_consider);
+      seconds_per_output = DIFFICULTY_TARGET_V2 * blocks_to_consider / outputs_to_consider;
```
vtnerd, line 66: Please flip back to the original logic (`outputs_per_second`), this removes a division in the `operator()` function.

src/util/gamma_picker.cpp
```diff
@@ -84,17 +89,27 @@ namespace lws
...
-      std::uint64_t output_index = std::exp(gamma(engine)) * outputs_per_second;
-      if (offsets().back() <= output_index)
+      double output_age_in_seconds = std::exp(gamma(engine));
+
+      // match wallet2 implementation to avoid creating distinct anonymity pools
+      // shift output back by unlock time to apply distribution from chain tip
+      if (output_age_in_seconds > defaut_unlock_time)
+        output_age_in_seconds -= defaut_unlock_time;
+      else
+        output_age_in_seconds = crypto::rand_idx(recent_spend_window);
```
vtnerd, line 103: I regret not looking closely at `wallet2`. This answer is hopefully in that review ... why this? Its arbitrary and ruins the exponential selection.

vtnerd (changes requested):

src/util/gamma_picker.cpp
vtnerd: Sorry let me restate -

Looks like I will have to go back over `wallet2` PR description more closely. I don't see how this `if` / `else` logic is correct, this is deviating from the source paper a bit.

vtnerd: Ok, just read the justification for the `recent_spend_window`. It probably makes sense given that the current `continue` creates a bias that doesn't follow the original intent of the source paper.

And I see what did you this logic, but I would like you to refactor. The existing version was already selecting from the chain tip, but this adds two subtractions (here and line 109) to match the code in `wallet2`. Trying to match that code led to an unnecessary `throw`.

Revert everything, and change the continue on the current line 97:
```c++
if (selection == end)
  output_index = *(offsets.rbegin() + crypto::rand_range(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE, std::min(50, offsets.size() - 1)))
static_assert(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE <= 50, "spendable must be less than spend window");
```

vtnerd: See Github for fixed code snippet

j-berman: On the subtraction: I figured it makes sense *not* to consider locked outputs when selecting outputs at all. Why should the outputs in that 10 block window be considered in the output line? There is no feasible way for a user to spend those outputs, so why should it be the case that they should have any impact at all on what outputs are likely to be spent by a user?  Imagine outputs spike in the 10 block window from the tip. This would cause the algorithm to select these outputs in greater frequency (and then redraw), but there is no reason for the algorithm to select those outputs in the first place because those outputs are not even feasibly spendable outputs. The general issue here is that the paper was written before the 10 block unlock time was enforced by consensus, so there needs to be some deviation to factor that in. This approach seemed to be the most sensible to me.

The above reasoning is why I stuck with moo's original approach where `num_rct_outputs` is the start of the output line where you ultimately subtract from, and went with the fairly odd (and admittedly code smelly) 2-step subtraction approach. 

On the throw: It's unnecessary because it should never happen. I just kept it there as a sanity check to signal the intent.

I think whatever we settle on here should be applied to wallet2 as well. So if we decide against the subtraction approach, we should probably make the change in wallet2 to match this implementation.

I should've asked you to review 7821 closely too, it would've been nice to have your eyes on it then as well and have this discussion then. I originally wanted to go with the exact same approach as yours (before I saw this code) -- your approach was the fix I initially proposed to moo when I initially reported it too. I wished the cleaner solution was also the best solution, but it didn't seem that way to me after sitting with it for a long time.

src/util/gamma_picker.cpp
j-berman: The point of this is to match integer truncation in wallet2 (https://github.com/monero-project/monero/pull/7845)

vtnerd: There doesn't seem to be any justification for integer truncation in that post. Just you claiming that two other PRs were more important. If anything, `wallet2` should be moving towards this implementation and not the other way.

src/util/gamma_picker.cpp
vtnerd: Random over `output_age_in_seconds` has a different distribution than random over `blocks` (I think that was your point).

In that case, rename `num_rct_outputs` to `num_unlocked_rct_outputs`. At first glance I thought this was shifting the distribution to the unlocked outputs. Also make `std::runtime_error` -> `std::logic_error` with some message indicating that it shouldn't happen (i.e. the logic needs to be looked at again).

src/util/gamma_picker.cpp
j-berman: > Also, I'm still serious about flipping that division logic. Not only is division more costly (barely), but seconds_per_output is less natural than outputs_per_second (at least I think so).

I much prefer the `seconds_per_output` approach as well, for both reasons too.

My comment explains that it's not necessarily a clear decision whether it is safe to patch integer truncation. I have (and had) more analysis that shows why it's not a clear decision -- I will share it here if you feel it's relevant. Continuing to pile onto that issue seemed to be overly confusing.

I agree that `wallet2` should ideally move towards removing integer truncation (and move toward this `outputs_per_second` approach), which is what I'm *also* continuing to work on fully assessing. Can just hold off on this PR until that work is complete if you'd like.

vtnerd: > I have (and had) more analysis that shows why it's not a clear decision -- I will share it here if you feel it's relevant. Continuing to pile onto that issue seemed to be overly confusing.

The primary argument I've seen is that you can't force an upgrade until the next hardfork  (uniformity issue). Is there another? Because I think you are making the integer truncation change too complicated -

MyMonero/Monero-LWS (the former representing a decent percentage of the transactions), didn't have the integer truncation issue (or the incorrect curve fit), so there's already a decent number of users with both changes. Which means there's already a lack of uniformity in the integer truncation case. And there's low probability of a benefit to an unintended truncation (it was unintended and accidentally made it better?).

vtnerd: I'm not seeing a compelling reason to accept the integer truncation issue (at least not yet). I have several questions, and possible criticisms of this analysis, which I am moving to the referenced Github issue.

This implementation was/is `wallet2` users, and now you want me to _under select_ new outputs, when you've done no analysis on **real** (as opposed to decoy) outputs.

j-berman: > This implementation was/is protecting wallet2 users, and now you want me to under select new outputs

I think it's definitely reasonable to not accept on these grounds. I figured it was desirable to have uniform implementations across wallet2 users and non-wallet2 users to ensure tx uniformity, but hearing your position I would agree that isn't absolutely critical or a strong reason enough in and of its own, without further evidence, to move away from integer truncation here.

> when you've done no analysis on real (as opposed to decoy) outputs.

The blue spikes = analysis on real outputs, but the crux of the position (to leave integer truncation unpatched) isn't well defended by those blue spikes.

When I said:

> I agree that wallet2 should ideally move towards removing integer truncation (and move toward this outputs_per_second approach), which is what I'm also continuing to work on fully assessing. Can just hold off on this PR until that work is complete if you'd like.

I will reword:

> I agree, the safer thing to do would be to just keep this PR on hold until further analysis on **real** outputs is complete.

This is what I intended to spend time working on.

j-berman: FWIW, I think the following chart from [this comment](https://github.com/monero-project/research-lab/issues/86#issuecomment-915806414):

mymonero recent shift from 2 to 1

is the visualized reason *not* to introduce integer truncation into monero-lws without more compelling evidence. Evidently this PR as is would introduce a larger under-selection for recent outputs (since green dash is below blue - - and red line is above blue -- for most of the range). And the cost of introducing that under-selection should be weighed against the benefit of higher tx uniformity we get from consistent implementations across wallet2/non-wallet2 users.

In hindsight, I shouldn't have tried to bring truncation to monero-lws without stronger reasoning. Your critical eye is appreciated.

Rucknium: @vtnerd I just want to chime in here and give some context. If all goes according to plan, this integer truncation issue will be moot in a few months since hopefully we will be moving to a brand new parametric distribution for the mixin selection algorithm.

The original suggestion of [Moser et al. (2018)](https://www.sciendo.com/article/10.1515/popets-2018-0025), which the current code attempts to implement, was a decent first draft, but in my view it never should have been implemented in production code in 2018, much less persisted until 2021. I base this view on my extensive training and experience in statistical analysis as an empirical microeconomist as well as Monero-specific theoretical and empirical investigations that I, @j-berman , and others have carried out recently. I am developing a CCS proposal to fund the research required to determine the best way forward in overhauling the mixin selection algorithm.

I am not sure about the merits of going with one fix or another with respect to the integer truncation issue. All that I wanted to say is that any fix will likely be quite temporary.

j-berman:
- Reverted back to `outputs_per_second` (removed the introduced integer truncation)
- Changed the runtime error to a logic error

Another comment: I proposed changing the `recent_spend_window` in wallet2 to 10 instead of 50 in [this comment](https://github.com/monero-project/monero/pull/7798#issuecomment-917495021) for next release. Figure you'd have thoughts on that as well

PR #29, opened 2022-02-04
First pass at informal audit

j-berman:
It's not complete, but I figure if all looks well to you, this can perhaps serve as a useful document to people coming to check out `monero-lws`, in addition to the other docs in the repo. If you don't agree, no worries :)

I didn't get as far with it as I'd have liked (partially because I lost my original file when my drive crashed, other part is that it was taking longer than I thought). But figure it's worth sharing as is anyway. It's divided into 3 parts:

1. How does `monero-lws` work?
2. The code
3. Suggestions

If people are interested, I can work to finish and fill it out here and there.


docs/review_02.03.22/review_02.03.22.md
```diff
@@ -0,0 +1,546 @@
...
+
+#### /src/util
+
+#### /src/wire
+
+### /CMakeLists.txt
+
+## Suggestions
+
+- Consider indexing addresses by a hash of the private view key instead of just
```
vtnerd, line 515: This is a decent suggestion. I _think_ the check for view_key -> view_public_key is always done before any DB lookups to prevent an obvious leak (that was my intent, but probably wasn't documented in the comments). Since LMDB is doing a sorted lookup, there might some issues with approved lws users recovering some lower order bits. The timing should depend on the number of accounts, the state of the file cache, state of the cpu cache, etc., but might be possible in some situations.

For those unable to follow - the recovery process should involve trying to determine where _your_ account is listed in the LMDB B-tree, which gives indirect information on the number of accounts and possible view-public key values of those accounts. Hashing would only "blind" the second case (the public-key values), and only when the attacker is attempting to recover values from nothing (i.e. if the attacker has some suspected account addresses, then the attacker can still hash+recover).

This is very pessimistic and must be weighed against having the first DB upgrade code in `monero-lws`. And if deemed important enough, its worth digging deeper to make other DB changes simultaneously to reduce amount of upgrade code.

vtnerd (commented):
These comments are mostly informational and intended to be commentary for anyone following `monero-lws` (as opposed to @j-berman and myself having some private IRC conversation).

I am holding this from merge (briefly) so that the fee field in the API can hopefully be corrected, and the corresponding comment in this PR immediately removed. A bit overkill, but at a glance the changes (in API code and this PR) should be easy to make.

There might also be one typo - was "SAFU" intentional? I can never tell anymore with crypto-currencies ;)

docs/review_02.03.22/review_02.03.22.md
```diff
@@ -0,0 +1,546 @@
...
+user's address and view key, and perpetually scanning for transactions received
+and plausibly spent by the user.
+
+The primary risk to a user using `monero-lws` as their backend is that the
+device where it runs is an additional target for someone malicious to compromise
+a user's privacy. A user should assume that someone with access to wherever the 
+`monero-lws` backend is hosted will be able to determine all outputs received to
+the user, and in most cases today, the outputs spent by the user as well. This
+also includes **amounts** received to and spent by the user. If a `monero-lws` 
+instance is compromised, user funds are still SAFU since the private spend key
```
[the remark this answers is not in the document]
j-berman: https://www.urbandictionary.com/define.php?term=Safu

It *is* an informal audit after all

docs/review_02.03.22/review_02.03.22.md
j-berman: > Since LMDB is doing a sorted lookup, there might some issues with approved lws users recovering some lower order bits.

I was debating clarifying that this is what I meant. Just some amount of the earlier bits could be revealed in theory.

> This is very pessimistic and must be weighed against having the first DB upgrade code in monero-lws

Agreed, very well could be not worth implementing.

docs/review_02.03.22/review_02.03.22.md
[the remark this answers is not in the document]
vtnerd: I don't there are multiple conflicting definitions there, and I haven't lurked on bitcointalk nearly long enough to know how the cool kids are using it.

docs/review_02.03.22/review_02.03.22.md
vtnerd: I intend to switch to a constant time comparison (in the specified LMDB sort) for the account address. This doesn't require any changes to the stored data, and helps hide the public view key during searching.

I am also leaning towards `cryptographic_hash(salt || public_view_key)` (in addition to the above), but I haven't thought through all of the scenarios yet.

vtnerd (approved):

Issue #7, opened 2020-10-08
boost: undefined reference


trasherdk:
I can't figure this one out. I'm getting this error:
```
/path/monero/monero-build/src/blockchain_db/libblockchain_db.a(db_lmdb.cpp.o): In function `boost::filesystem::unique_path(boost::filesystem::path const&)':
db_lmdb.cpp:(.text._ZN5boost10filesystem11unique_pathERKNS0_4pathE[_ZN5boost10filesystem11unique_pathERKNS0_4pathE]+0x33): undefined reference to `boost::filesystem::detail::unique_path(boost::filesystem::path const&, boost::system::error_code*)'
```
My system boost library is `boost-1.59.0` and the Monero one seems to be `boost-1.64.0` but the declarations are identical.

I have checked boost header file filesystem/operations.hpp and both function are defined.
```
/usr/include/boost/filesystem/operations.hpp
    BOOST_FILESYSTEM_DECL
    path unique_path(const path& p, system::error_code* ec=0);

  inline
  path unique_path(const path& p="%%%%-%%%%-%%%%-%%%%")
                                       { return detail::unique_path(p); }
  inline
  path unique_path(const path& p, system::error_code& ec)
                                       { return detail::unique_path(p, &ec); }
```

I checked the library file
```
/usr/lib64/libboost_filesystem.a
 _ZN5boost10filesystem6detail11unique_pathERKNS0_4pathEPNS_6system10error_codeE
```
And the one in the error
```
From the error
 _ZN5boost10filesystem11unique_pathERKNS0_4pathE
```
They seem to point to different functions ???

So. What am I missing here? What can I do?

vtnerd:
do `message("${Boost_INCLUDE_DIR}")` somewhere after the boost find step to see which Boost version is being used. This seems like a mismatch between header-file and library file.

trasherdk:
They looks fine to me:
```
Boost_INCLUDE_DIR.........: /usr/include
Boost_THREAD_LIBRARY......: /usr/lib64/libboost_thread.a
Boost_FILESYSTEM_LIBRARY..: /usr/lib64/libboost_filesystem.a
```

Is Monero build against something else? I can't find `boost/filesystem/*` with `unique_path` .

trasherdk:
It seems that `monero-lws-daemon` does not like a static linked version of boost.

The `monero-lws-admin` did build against a static boost without problems.

I removed the static boost, installed a dynamic, same version `1.59.0`, and the build completed.

[...]

PR #30, opened 2022-02-08
Adding a Dockerfile

j-berman:
Requested on [the reddit](https://www.reddit.com/r/Monero/comments/sll4bi/monero_meet_20220205_fees_monerolws_and_network/hvw1zzw/)

### Instructions

If you're already running `monerod` on the same machine you'd like to host `monero-lws`, you can use this Dockerfile to spin up a `monero-lws` docker container by running the following commands on Linux:

```
docker build -t monero-lws .

# starts monero-lws-daemon
docker run --network="host" --restart unless-stopped --name="monero-lws" -v bitmonero-lws:/home/monero-lws/.bitmonero/light_wallet_server monero-lws --log-level=4

# run admin commands as described here: https://github.com/vtnerd/monero-lws/blob/develop/docs/administration.md
docker exec monero-lws monero-lws-admin --db-path="home/monero-lws/.bitmonero/light_wallet_server" --help
docker exec monero-lws monero-lws-admin --db-path="home/monero-lws/.bitmonero/light_wallet_server" debug_database | json_pp
```

On Linux, `--network="host"` will expose the host machine's network to the Docker container. This is what allows the Docker container to retrieve data from the running `monerod` instance on the machine. More on this [here](https://docs.docker.com/network/network-tutorial-host/).

On [Windows](https://docs.docker.com/desktop/windows/networking/#use-cases-and-workarounds) and [Mac](https://docs.docker.com/desktop/mac/networking/#use-cases-and-workarounds), the `docker run` command I believe should be this instead (untested):

```
docker run --restart unless-stopped --name="monero-lws" -v bitmonero-lws:/home/monero-lws/.bitmonero/light_wallet_server monero-lws --log-level=4 --daemon="tcp://host.docker.internal:18082"
```

There should be a way to write [a consistent docker run command](https://stackoverflow.com/a/43541732) for all platforms as of Docker v20.10, but couldn't get it working so moving on. Certain there's room for cleanup in these commands by docker afficionados.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,19 +88,36 @@ if (NOT MONERO_BUILD_DIR)
   message(FATAL_ERROR "The argument -DMONERO_BUILD_DIR must specify a location of an existing monero build")
 endif()
 
+if(STATIC)
+  if(MSVC)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .lib .dll.a .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+  else()
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+  endif()
+endif()
+
 load_cache(${MONERO_BUILD_DIR} READ_WITH_PREFIX monero_
   Boost_THREAD_LIBRARY_RELEASE
   CMAKE_CXX_COMPILER
   EXTRA_LIBRARIES
+  LIBUDEV_LIBRARY
+  usb_LIBRARY
   HIDAPI_INCLUDE_DIR
+  HIDAPI_LIBRARY
   LMDB_INCLUDE
   monero_SOURCE_DIR
+  OPENSSL_INCLUDE_PATH
   OPENSSL_CRYPTO_LIBRARY
   OPENSSL_SSL_LIBRARY
   SODIUM_LIBRARY
   UNBOUND_LIBRARIES
   ZMQ_INCLUDE_PATH
   ZMQ_LIB
+  PGM_LIBRARY
+  NORM_LIBRARY
+  GSSAPI_LIBRARY
+  PROTOLIB_LIBRARY
+  SODIUM_LIBRARY
 )
 
 if (NOT (monero_monero_SOURCE_DIR MATCHES "${MONERO_SOURCE_DIR}(/src/cryptonote_protocol)"))
@@ -136,101 +153,34 @@ if (NOT (Boost_THREAD_LIBRARY STREQUAL monero_Boost_THREAD_LIBRARY_RELEASE))
   message(FATAL_ERROR "Boost libraries for monero build differs from this project")
 endif()
 
-# hidapi
-# https://github.com/moneroexamples/openmonero/blob/18370af8fa51a0e9d05dfe678dcd5bcb7fad30f7/cmake/FindHIDAPI.cmake#L26-L60
-find_library(HIDAPI_LIBRARY
-  NAMES hidapi hidapi-libusb)
 
-find_path(HIDAPI_INCLUDE_DIR
-  NAMES hidapi.h
-  PATH_SUFFIXES
-  hidapi)
+set(LMDB_INCLUDE "${monero_LMDB_INCLUDE}")
+set(LMDB_LIB_PATH "monero::lmdb")
+set(ZMQ_LIB "${monero_ZMQ_LIB}")
 
-include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(HIDAPI
-  DEFAULT_MSG
-  HIDAPI_LIBRARY
-  HIDAPI_INCLUDE_DIR)
-
-if(HIDAPI_FOUND)
-  set(HIDAPI_LIBRARIES "${HIDAPI_LIBRARY}")
-  if((STATIC AND UNIX AND NOT APPLE) OR (DEPENDS AND CMAKE_SYSTEM_NAME STREQUAL "Linux"))
-    find_library(LIBUSB-1.0_LIBRARY usb-1.0)
-    find_library(LIBUDEV_LIBRARY udev)
-    if(LIBUSB-1.0_LIBRARY)
-      set(HIDAPI_LIBRARIES "${HIDAPI_LIBRARIES};${LIBUSB-1.0_LIBRARY}")
-      if(LIBUDEV_LIBRARY)
-        set(HIDAPI_LIBRARIES "${HIDAPI_LIBRARIES};${LIBUDEV_LIBRARY}")
-      else()
-        message(WARNING "libudev library not found, binaries may fail to link.")
-      endif()
-    else()
-      message(WARNING "libusb-1.0 library not found, binaries may fail to link.")
-    endif()
-  endif()
-
-  set(HIDAPI_INCLUDE_DIRS "${HIDAPI_INCLUDE_DIR}")
-endif()
-
-mark_as_advanced(HIDAPI_INCLUDE_DIR HIDAPI_LIBRARY)
-
-if (NOT (HIDAPI_INCLUDE_DIR STREQUAL monero_HIDAPI_INCLUDE_DIR))
-  message(STATUS "Found HIDAPI_INCLUDE_DIR: ${HIDAPI_INCLUDE_DIR}")
-  message(STATUS "Found monero_HIDAPI_INCLUDE_DIR: ${monero_HIDAPI_INCLUDE_DIR}")
-  message(FATAL_ERROR "hidapi libraries for monero build differs from this project")
-endif()
-
-# openssl
-if (STATIC)
-  set(OPENSSL_USE_STATIC_LIBS TRUE)
-endif()
-find_package(OpenSSL REQUIRED)
-
-if (NOT (OPENSSL_CRYPTO_LIBRARY STREQUAL monero_OPENSSL_CRYPTO_LIBRARY))
-  message(STATUS "Found OPENSSL_CRYPTO_LIBRARY: ${OPENSSL_CRYPTO_LIBRARY}")
-  message(STATUS "Found monero_OPENSSL_CRYPTO_LIBRARY: ${monero_OPENSSL_CRYPTO_LIBRARY}")
-  message(FATAL_ERROR "openssl libraries for monero build differs from this project")
-endif()
-
-# zmq
-# https://github.com/monero-project/monero/blob/9aab19f349433687c7aaf2c1cbc5751e5912c0aa/CMakeLists.txt#L1171-L1199
-if(STATIC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DZMQ_STATIC")
-endif()
-find_path(ZMQ_INCLUDE_PATH zmq.h)
-find_library(ZMQ_LIB zmq)
-find_library(PGM_LIBRARY pgm)
-find_library(NORM_LIBRARY norm)
-find_library(GSSAPI_LIBRARY gssapi_krb5)
-find_library(PROTOLIB_LIBRARY protolib)
-find_library(SODIUM_LIBRARY sodium)
-
-if(NOT ZMQ_INCLUDE_PATH)
-  message(FATAL_ERROR "Could not find required header zmq.h")
-endif()
 if(NOT ZMQ_LIB)
   message(FATAL_ERROR "Could not find required libzmq")
 endif()
-if(PGM_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${PGM_LIBRARY}")
+if(monero_PGM_LIBRARY)
+  set(ZMQ_LIB "${ZMQ_LIB};${monero_PGM_LIBRARY}")
 endif()
-if(NORM_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${NORM_LIBRARY}")
+if(monero_NORM_LIBRARY)
+  set(ZMQ_LIB "${ZMQ_LIB};${monero_NORM_LIBRARY}")
 endif()
-if(GSSAPI_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${GSSAPI_LIBRARY}")
+if(monero_GSSAPI_LIBRARY)
+  set(ZMQ_LIB "${ZMQ_LIB};${monero_GSSAPI_LIBRARY}")
 endif()
-if(PROTOLIB_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${PROTOLIB_LIBRARY}")
+if(monero_PROTOLIB_LIBRARY)
+  set(ZMQ_LIB "${ZMQ_LIB};${monero_PROTOLIB_LIBRARY}")
 endif()
-if(SODIUM_LIBRARY)
-  set(ZMQ_LIB "${ZMQ_LIB};${SODIUM_LIBRARY}")
+if(monero_SODIUM_LIBRARY)
+  set(ZMQ_LIB "${ZMQ_LIB};${monero_SODIUM_LIBRARY}")
 endif()
 
-if (NOT (ZMQ_INCLUDE_PATH STREQUAL monero_ZMQ_INCLUDE_PATH))
-  message(STATUS "Found ZMQ_INCLUDE_PATH: ${ZMQ_INCLUDE_PATH}")
-  message(STATUS "Found monero_ZMQ_INCLUDE_PATH: ${monero_ZMQ_INCLUDE_PATH}")
-  message(FATAL_ERROR "zmq libraries for monero build differs from this project")
+if(STATIC AND NOT IOS)
+  if(UNIX)
+    set(monero_OPENSSL_LIBRARIES "${monero_OPENSSL_LIBRARIES};${CMAKE_DL_LIBS};${CMAKE_THREAD_LIBS_INIT}")
+  endif()
 endif()
 
 foreach (LIB ${MONERO_LIBRARIES})
@@ -247,7 +197,6 @@ foreach (LIB ${MONERO_LIBRARIES})
     set_target_properties(${LIB_NAME} PROPERTIES IMPORTED_LOCATION ${LIB_PATH})
     list(APPEND IMPORTED_MONERO_LIBRARIES "${LIB_NAME}")
   endif()
-  
   unset(LIB_PATH CACHE)
 endforeach()
 
@@ -255,9 +204,8 @@ add_library(monero::libraries INTERFACE IMPORTED)
 set_property(TARGET monero::libraries PROPERTY
   INTERFACE_INCLUDE_DIRECTORIES
     ${Boost_INCLUDE_DIR}
-    ${HIDAPI_INCLUDE_DIRS}
-    ${OPENSSL_INCLUDE_DIR}
-    ${ZMQ_INCLUDE_PATH}
+    ${monero_HIDAPI_INCLUDE_DIRS}
+    ${monero_OPENSSL_INCLUDE_PATH}
     "${MONERO_BUILD_DIR}/generated_include"
     "${MONERO_SOURCE_DIR}/contrib/epee/include"
     "${MONERO_SOURCE_DIR}/external/easylogging++"
@@ -274,16 +222,16 @@ set_property(TARGET monero::libraries PROPERTY
     ${Boost_SERIALIZATION_LIBRARY}
     ${Boost_SYSTEM_LIBRARY}
     ${Boost_THREAD_LIBRARY}
-    ${HIDAPI_LIBRARIES}
-    ${OPENSSL_LIBRARIES}
+    ${monero_HIDAPI_LIBRARY}
+    ${monero_usb_LIBRARY}
+    ${monero_LIBUDEV_LIBRARY}
+    ${ZMQ_LIB}
+    ${monero_OPENSSL_SSL_LIBRARY}
+    ${monero_OPENSSL_CRYPTO_LIBRARY}
     ${monero_SODIUM_LIBRARY}
     ${monero_UNBOUND_LIBRARIES}
     ${CMAKE_DL_LIBS}
 )
-
-set(LMDB_INCLUDE "${monero_LMDB_INCLUDE}")
-set(LMDB_LIB_PATH "monero::lmdb")
-
 
 #
 # Build monero-lws code
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,7 @@ load_cache(${MONERO_BUILD_DIR} READ_WITH_PREFIX monero_
   Boost_THREAD_LIBRARY_RELEASE
   CMAKE_CXX_COMPILER
   EXTRA_LIBRARIES
-  HIDAPI_LIBRARY
+  HIDAPI_INCLUDE_DIR
   LMDB_INCLUDE
   monero_SOURCE_DIR
   OPENSSL_CRYPTO_LIBRARY
@@ -115,10 +115,6 @@ if ("${monero_UNBOUND_LIBRARIES}" STREQUAL "UNBOUND_LIBRARIES-NOTFOUND")
   unset(monero_UNBOUND_LIBRARIES)
 endif()
 
-if ("${monero_HIDAPI_LIBRARY}" STREQUAL "HIDAPI_LIBRARY-NOTFOUND")
-  unset(monero_HIDAPI_LIBRARY)
-endif()
-
 #
 # Dependencies specific to monero-lws
 #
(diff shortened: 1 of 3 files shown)


CryptoGrampy:
@j-berman I have no clue what I'm doing (smooth brain here), but I'm trying to wire this up to an existing monerod service with docker compose. 
I'm able to get it to start up, but I have no idea what the startup commands would be /what ports need to be linked with and exposed in monerod.  It seems like 18082 may need to be added along with 18083?  And I'd also like to be able to say `--some-port:my-monerod-service:18082`.  Not sure if that's possible with the current startup flags.  Let me know if you have any feedback.  The 192-XX stuff is just so I can point a wallet at the server to test. 

Other stuff- it's a *very* large image- about 2.5 GB

CryptoGrampy:
Okay, was able to get this working with Seth's monerod docker image and traefik/dns provider using (not including a few items like traefik/dns service setup)

`---
version: "3.5"
services:
    monero-lws:
        // manually build and tag monero-lws:latest from jberman's dockerfile first
        image: monero-lws:latest
        restart: unless-stopped
        volumes:
        - monero-lws:/home/monero-lws/.bitmonero/light_wallet_server
        command: >-
          --daemon=tcp://my-server-ip:18082
          --sub=tcp://my-server-ip:18083
          --log-level=4
          --confirm-external-bind
          --rest-server=http://0.0.0.0:80
        labels:
        - traefik.enable=true
        - traefik.http.services.monero-lws.loadbalancer.server.port=80
        - traefik.http.routers.monero-lws.rule=Host(`lws.myserver.com`)
        - traefik.http.routers.monero-lws.tls.certresolver=le
        - traefik.http.routers.monero-lws.entrypoints=websecure
    monerod:
        image: sethsimmons/simple-monerod:alpine
        restart: unless-stopped
        container_name: monerod
        volumes:
        - bitmonero:/home/monero
        ports:
        - 18080:18080
        - 127.0.0.1:18081:18081
        - 18083:18083
        - 18082:18082
        - 18089:18089
        command: >-
            --rpc-restricted-bind-ip=0.0.0.0 
            --rpc-restricted-bind-port=18089
            --rpc-bind-ip=0.0.0.0
            --rpc-bind-port=18081
            --confirm-external-bind
            --no-igd
            --enable-dns-blocklist
            --prune-blockchain
            --zmq-pub=tcp://0.0.0.0:18083 
            --zmq-rpc-bind-ip=0.0.0.0
            --zmq-rpc-bind-port=18082
            --in-peers=10
            --out-peers=10
volumes:
    bitmonero:
    monero-lws:

`
@j-berman I think it would be good to add jq to the image, then you shell into the container using:
`docker exec -it  monero-lws /bin/bash`

and then from within the container: `monero-lws-admin accept_requests create $(monero-lws-admin list_requests | jq -j '.create? | .[]? | .address?+" "')` etc.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    && cmake -DMONERO_SOURCE_DIR=/monero -DMONERO_BUILD_DIR=/monero/build/Linux/_no_branch_/release .. \
+    && make -j$(nproc)
+
+# expose REST server port
+EXPOSE 8443
+
+# expose commands `monero-lws-daemon` and `monero-lws-admin` via PATH
+ENV PATH "$PATH:/monero-lws/build/src"
+
+ENTRYPOINT ["monero-lws-daemon"]
```
CryptoGrampy, line 40: Data persistence issue- I'm not a Docker pro, but I think we want to add something like this after line 40 because the container is not actually persisting data when you specify a volume on the host/container (as specified in your startup command)

`CMD ["--db-path=/home/monero-lws/.bitmonero/light_wallet_server"]`

Basically, if you don't add this line as part of your docker/docker-compose startup command, i think the container defaults to putting the .mdb files in some root access directory and they don't persist in the specified docker volume.  I think adding the CMD line makes more sense than telling people to specify it manually.

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+ARG MONERO_LWS_BRANCH=release-v0.1_0.17
+ARG MONERO_LWS_COMMIT_HASH=9d8a7e558aace21782217f69c3f083069538f38a
+
+# install dependencies
+RUN set -o errexit -o nounset \
+    && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get update \
+    && apt-get upgrade -y -q \
+    && apt-get install -y -q git build-essential cmake pkg-config libssl-dev libzmq3-dev libunbound-dev libsodium-dev libunwind8-dev \
+    liblzma-dev libreadline6-dev libldns-dev libexpat1-dev libpgm-dev qttools5-dev-tools libhidapi-dev libusb-1.0-0-dev \
```
CryptoGrampy, line 16: add `jq`

j-berman:
@CryptoGrampy thank you for going further with this :D ACK to all your comments, will take another pass soon.

On traefik, I've never used it before but with a quick pass, it looks like the architecture you've got set up is an internet-facing load balancer which functions as a proxy to a private network, where `monero-lws` + `monerod` reside. It looks like you have TLS configured on the load balancer, so https requests are supported, which is good. And then you expose `monero-lws` to the load balancer over http, which is fine because I assume it's on a private network.

2 important notes:
(1) are you certain that `monero-lws` is not directly exposed to the internet in your setup?
(2) we also need to double check that MyMonero clients are making initial requests over https, [this issue](https://github.com/mymonero/mymonero-android-js/issues/75) gives me pause

If hazy on the above, I think it would be good to include a disclaimer that for people trying [this setup](https://github.com/vtnerd/monero-lws/pull/30#issuecomment-1033176520) out, they should only use test view key until we have definitive answers. I'll be able to help out with this soon, but while we might see some traffic until we have solid answers, I think a disclaimer would be good to include (unless you've got the answers already :D )

CryptoGrampy:
If anyone is able to get this running through reading the full PR/comments, I hope they don't put their life savings on it :-) That would most certainly not be a good idea, so anyone reading this... we're TESTING here! 

But you're correct- I have Traefik set up as a reverse proxy on a home server- it intercepts any external network traffic on 443/80 and sends it where it needs to go (just a few docker services).  I believe with my the current setup, the only way to access the LWS server is through the lws.server.com that i've specified and any connection attempt I'm making over :80 seems to be getting upgraded to :443 in the browser with the proper LE cert.  I don't seem to be able to connect to LWS from within my LAN with this config either.  I am also connecting to the server from outside my network without any hiccups.  I'll pull MyMonero and attempt running it in dev mode and watch the network traffic (initial requests in particular) to LWS.  

BTW, I'm running a setup very similar to Seth: https://github.com/sethforprivacy/self-hosted-services/blob/main/docker-compose.yml

CryptoGrampy:
Just want to confirm (but I could certainly be wrong!) that when I specify lws.myserver.com in the MyMonero app on Linux Desktop and then create a new wallet, the first post request that contains `{ address: "xxxxx", view_key: "xxxxxxx"` is sent to the lws server /login endpoint via https and manually specifying httP://lws.myserver.com in MyMonero config still results in requests being sent over httpS.  I don't see any user data or addresses popping up in my Traefik logs, and the TLS stuff appears to be working correctly if I send the same login requests and post body in Postman.

vtnerd:
> Just want to confirm (but I could certainly be wrong!) that when I specify lws.myserver.com in the MyMonero app on Linux Desktop and then create a new wallet, the first post request that contains { address: "xxxxx", view_key: "xxxxxxx" is sent to the lws server /login endpoint via https and manually specifying http://lws.myserver.com/ in MyMonero config still results in requests being sent over httpS. I don't see any user data or addresses popping up in my Traefik logs, and the TLS stuff appears to be working correctly if I send the same login requests and post body in Postman.

What wallet are you using? It looks like MyMonero? There was some funkiness with that wallet and https, but I haven't checked the last release (or few) to how the situation changed.

Dockerfile
vtnerd: `jq` is _not_ needed to run, but is useful for the bash commands listed in the documentation. I think it makes more sense to omit it as a default installed component, but I am also less familiar with docker norms.

j-berman:
I'm gonna double check the code for all different MyMonero clients to see what they're doing when I get the chance

CryptoGrampy:
> What wallet are you using? It looks like MyMonero? There was some funkiness with that wallet and https, but I haven't checked the last release (or few) to see how the situation changed.

Using the latest 1.2.4 MyMonero AppImage on Linux

CryptoGrampy:
> I'm gonna double check the code for all different MyMonero clients (whatever's in the latest [Android](https://github.com/mymonero/mymonero-android-js), [iOS](https://github.com/mymonero/mymonero-app-ios) and [desktop](https://github.com/mymonero/mymonero-app-js) repos) to see what they're doing when I get the chance

https://github.com/mymonero/mymonero-utils/blob/583cfa1a5b18f3080c1278f1f615e3a886822d93/packages/mymonero-net-service-utils/index.js#L59

Dockerfile
CryptoGrampy: I was thinking a good next step for this PR would be basing the image on Alpine to reduce size and complexity, as well as moving to a multi-step build where the built binaries (admin and the daemon?) are essentially copy/pasted into a subsequent build process that only adds 'necessary' packages for running and using these binaries- this is where I would probably put jq... 

I'm not super familiar with LWS, but if there aren't any RPC methods for admin stuff and jq is the easiest current way to do these maintenance tasks (it appears to be the documented way), then I would probably support adding it in.  Otherwise, everyone that reads the current documentation will have to manually install it within the running container every time it restarts...  and they probably will do that. 

There's a P2Pool Dockerfile that basically does this exact process (including adding jq at the end): https://github.com/WeebDataHoarder/p2pool-compose/blob/master/docker/p2pool/Dockerfile in case you wanto to check it out.

sethforprivacy:
I have some ideas on how to make this a multi-stage build which should reduce the final image size *drastically*.

It will require basically a total rewrite, but in the meantime the Dockerfile looks solid (pending me having a chance to test it personally.

Would you rather I make recommendations for a multi-stage build here, or wait and open a PR once this is merged?

Dockerfile
sethforprivacy: I'd definitely recommend adding jq in build if it's at all necessary to normal operations and admin functions for LWS (which it seems to be), as installing packages after build is usually quite difficult as the running user in the container should *not* be root or have sudo permissions.

j-berman:
> It will require basically a total rewrite, but in the meantime the Dockerfile looks like a great start (pending me having a chance to test it personally).

@sethforprivacy I'm thinking may as well wait to put something out until you get a go at it. Considering people will have persistent volumes stored somewhere as a result of whatever gets merged into the repo, it makes sense to try and get the volume stored in the optimal spot in the right way from the get-go. Can feel free to just submit a new PR starting from scratch, nbd :)

sethforprivacy:
> > It will require basically a total rewrite, but in the meantime the Dockerfile looks like a great start (pending me having a chance to test it personally).
> 
> @sethforprivacy I'm thinking may as well wait to put something out until you get a go at it. Considering people will have persistent volumes stored somewhere as a result of whatever gets merged into the repo, it makes sense to try and get the volume stored in the optimal spot in the right way from the get-go. Can feel free to just submit a new PR starting from scratch, nbd :)

This is an *amazing* base and knocks out all of the hard work for the Dockerfile, thankful I don't need to start from scratch now!

I'll copy-paste from here, give credit, and open a new PR after some testing later today. Thanks so much for the work on this, @j-berman!

sethforprivacy:
Initial take on a more optimized Dockerfile is here: https://github.com/sethforprivacy/monero-lws/blob/develop/Dockerfile

This is probably as small and clean as we can make it until static compilation is possible for monero-lws. I reached out to @vtnerd and @j-berman about static compilation directly, but that's the biggest missing piece for a better Dockerfile at this point.

@j-berman -- feel free to either snag what you like from mine and roll it into your PR (I don't care about attribution to mine) or I can open a PR after I test mine a bit more.

Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+    && test `git rev-parse HEAD` = ${MONERO_COMMIT_HASH} || exit 1 \
+    && git submodule init && git submodule update \
+    && mkdir -p build/release && cd build/release \
+    # Create make build files manually for release-static-linux-x86_64
+    && cmake -D STATIC=ON -D ARCH="x86-64" -D BUILD_64=ON -D CMAKE_BUILD_TYPE=release -D BUILD_TAG="linux-x64" ../.. \
+    # Build only monerod binary using number of available threads
+    && cd /monero && nice -n 19 ionice -c2 -n7 make -j${NPROC:-$(nproc)} -C build/release daemon
+
+# TODO: remove the need to manually make this static liblmdb_lib.a
+RUN cd /monero/build/release/src/lmdb && make && cd /monero
```
j-berman, line 69: Fixes:

```
CMake Error at CMakeLists.txt:242 (message):
  Unable to find required Monero library lmdb_lib
```

src/CMakeLists.txt
```diff
@@ -53,8 +53,8 @@ target_link_libraries(monero-lws-daemon
     monero-lws-util
     monero-lws-wire-json
     ${Boost_CHRONO_LIBRARY}
-    ${Boost_FILESYSTEM_LIBRARY}
     ${Boost_PROGRAM_OPTIONS_LIBRARY}
+    ${Boost_FILESYSTEM_LIBRARY}
```
j-berman, line 57: Fixes #7 :)

Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+# Install necessary dependencies
+RUN apt-get install --no-install-recommends -y \
+    ca-certificates \
+    curl \
+    jq \
+    libgssapi-krb5-2 \
+    libhidapi-dev \
+    libnorm-dev \
+    libpgm-dev \
+    libzmq3-dev \
```
j-berman, line 103: Fixes:

```
monero-lws-daemon: error while loading shared libraries: libzmq.so.5: cannot open shared object file: No such file or directory
```

Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+RUN apt-get update \
+    && apt-get upgrade --no-install-recommends -y
+
+# Install necessary dependencies
+RUN apt-get install --no-install-recommends -y \
+    ca-certificates \
+    curl \
+    jq \
+    libgssapi-krb5-2 \
+    libhidapi-dev \
```
j-berman, line 100: Fixes:

```
monero-lws-daemon: error while loading shared libraries: libhidapi-libusb.so.0: cannot open shared object file: No such file or directory
```

CMakeLists.txt
```diff
@@ -173,18 +274,15 @@ set_property(TARGET monero::libraries PROPERTY
     ${Boost_SERIALIZATION_LIBRARY}
     ${Boost_SYSTEM_LIBRARY}
     ${Boost_THREAD_LIBRARY}
-    ${monero_HIDAPI_LIBRARY}
-    ${monero_OPENSSL_CRYPTO_LIBRARY}
-    ${monero_OPENSSL_SSL_LIBRARY}
+    ${HIDAPI_LIBRARIES}
+    ${OPENSSL_LIBRARIES}
```
j-berman, line 278: `${OPENSSL_LIBRARIES}` fixes compiler errors like:

```
/usr/bin/ld: /monero/build/release/contrib/epee/src/libepee.a(net_ssl.cpp.o): in function `epee::net_utils::ssl_options_t::has_fingerprint(boost::asio::ssl::verify_context&) const':
net_ssl.cpp:(.text+0x3995): undefined reference to `X509_STORE_CTX_get0_chain'
/usr/bin/ld: net_ssl.cpp:(.text+0x39a5): undefined reference to `OPENSSL_sk_num'
/usr/bin/ld: net_ssl.cpp:(.text+0x3c66): undefined reference to `OPENSSL_sk_value'
/usr/bin/ld: net_ssl.cpp:(.text+0x3ccc): undefined reference to `EVP_sha256'
/usr/bin/ld: net_ssl.cpp:(.text+0x3cdf): undefined reference to `X509_digest'
```

CMakeLists.txt
```diff
@@ -135,6 +136,103 @@ if (NOT (Boost_THREAD_LIBRARY STREQUAL monero_Boost_THREAD_LIBRARY_RELEASE))
...
+find_package(OpenSSL REQUIRED)
+
+if (NOT (OPENSSL_CRYPTO_LIBRARY STREQUAL monero_OPENSSL_CRYPTO_LIBRARY))
+  message(STATUS "Found OPENSSL_CRYPTO_LIBRARY: ${OPENSSL_CRYPTO_LIBRARY}")
+  message(STATUS "Found monero_OPENSSL_CRYPTO_LIBRARY: ${monero_OPENSSL_CRYPTO_LIBRARY}")
+  message(FATAL_ERROR "openssl libraries for monero build differs from this project")
+endif()
+
+# zmq
+# https://github.com/monero-project/monero/blob/9aab19f349433687c7aaf2c1cbc5751e5912c0aa/CMakeLists.txt#L1171-L1199
```
j-berman, line 196: zmq changes in this file are here to fix compiler errors like:

```
/usr/bin/ld: /monero/build/release/src/net/libnet.a(zmq.cpp.o): in function `net::zmq::error_category()::category::message(int) const':
zmq.cpp:(.text+0xba): undefined reference to `zmq_strerror'
/usr/bin/ld: /monero/build/release/src/net/libnet.a(zmq.cpp.o): in function `net::zmq::terminate::call(void*)':
zmq.cpp:(.text+0x251): undefined reference to `zmq_errno'
/usr/bin/ld: zmq.cpp:(.text+0x25e): undefined reference to `zmq_term'
/usr/bin/ld: /monero/build/release/src/net/libnet.a(zmq.cpp.o): in function `net::zmq::receive[abi:cxx11](void*, int)':
zmq.cpp:(.text+0x2e4): undefined reference to `zmq_msg_init'
/usr/bin/ld: zmq.cpp:(.text+0x2f1): undefined reference to `zmq_msg_size'
/usr/bin/ld: zmq.cpp:(.text+0x2fe): undefined reference to `zmq_msg_data'
/usr/bin/ld: zmq.cpp:(.text+0x329): undefined reference to `zmq_msg_more'
/usr/bin/ld: zmq.cpp:(.text+0x33f): undefined reference to `zmq_msg_recv'
/usr/bin/ld: zmq.cpp:(.text+0x34b): undefined reference to `zmq_msg_close'
/usr/bin/ld: zmq.cpp:(.text+0x350): undefined reference to `zmq_errno'
/usr/bin/ld: zmq.cpp:(.text+0x3cc): undefined reference to `zmq_msg_close'
```

Pretty much just copy pasted the `monerod` code linked above.

CMakeLists.txt
```diff
@@ -135,6 +136,103 @@ if (NOT (Boost_THREAD_LIBRARY STREQUAL monero_Boost_THREAD_LIBRARY_RELEASE))
   message(FATAL_ERROR "Boost libraries for monero build differs from this project")
 endif()
 
+# hidapi
+# https://github.com/moneroexamples/openmonero/blob/18370af8fa51a0e9d05dfe678dcd5bcb7fad30f7/cmake/FindHIDAPI.cmake#L26-L60
```
j-berman, line 140: HIDAPI changes in this file are here to fix compiler errors like:

```
/usr/bin/ld: /monero/build/release/src/device/libdevice.a(device_io_hid.cpp.o): in function `hw::io::device_io_hid::disconnect() [clone .localalias]':
device_io_hid.cpp:(.text+0x32): undefined reference to `hid_close'
```

Pretty much just copy pasted the `openmonero` code linked above.

j-berman (commented):
Started from @sethforprivacy's [Dockerfile in the comment above](https://github.com/vtnerd/monero-lws/pull/30#issuecomment-1039595513), and hacked my way to linking a static build of `monerod` to `monero-lws`. I hope this can serve as a springboard for someone with stronger "make-fu" than me to see where things can be cut down/improved.

The final image size is 241mb on my machine.

Leaving comments on some areas I'm certain can probably be done cleaner and probably "more correctly" (i.e. where I "hacked").

sethforprivacy:
These changes look great, glad someone who actually understands make/cmake could figure this one out 😅 

Will give it a try as I have time.

CryptoGrampy:
For anyone testing this out, I'm already finding some bad privacy behavior with the MyMonero app- please just use this for testing only.  https://github.com/mymonero/mymonero-app-js/issues/458

sethforprivacy:
Another similarly bad bug on Android specifically -- https://github.com/mymonero/mymonero-android-js/issues/83

Only use testing wallets without funds when testing LWS until these serious issues are resolved.

sethforprivacy:
> Started from @sethforprivacy's [Dockerfile in the comment above](https://github.com/vtnerd/monero-lws/pull/30#issuecomment-1039595513), and hacked my way to linking a static build of `monerod` to `monero-lws`. I hope this can serve as a springboard for someone with stronger "make-fu" than me to see where things can be cut down/improved.
> 
> The final image size is 241mb on my machine.
> 
> Leaving comments on some areas I'm certain can probably be done cleaner and probably "more correctly" (i.e. where I "hacked").

Tested without issues on two separate hosts, looks *great* and greatly reduces the size of the image! Excellent work on this.

sethforprivacy:
Once this is done, tested, and merged I will PR Github Actions for automatically building and publishing Docker images using the Dockerfile.

Will pull heavily from [here](https://github.com/sethforprivacy/simple-monerod-docker/tree/main/.github/workflows) so that end-users won't have to build from source. 

If that is undesired @vtnerd let me know and I can keep a fork up that publishes under my own user instead.

j-berman:
Fees are still fingerprintable
MyMonero: https://github.com/mymonero/mymonero-core-cpp/pull/35, https://github.com/mymonero/mymonero-core-cpp/pull/36
lws: https://github.com/vtnerd/monero-lws/pull/31

CMakeLists.txt
```diff
@@ -135,6 +136,103 @@ if (NOT (Boost_THREAD_LIBRARY STREQUAL monero_Boost_THREAD_LIBRARY_RELEASE))
...
+  message(STATUS "Found monero_OPENSSL_CRYPTO_LIBRARY: ${monero_OPENSSL_CRYPTO_LIBRARY}")
+  message(FATAL_ERROR "openssl libraries for monero build differs from this project")
+endif()
+
+# zmq
+# https://github.com/monero-project/monero/blob/9aab19f349433687c7aaf2c1cbc5751e5912c0aa/CMakeLists.txt#L1171-L1199
+if(STATIC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DZMQ_STATIC")
+endif()
+find_path(ZMQ_INCLUDE_PATH zmq.h)
```
vtnerd, line 200: I don't understand these changes here, why not use the same stuff from the daemon? And why the extra libraries, are there linker errors in Ubuntu (this is reference to protolib, etc.)

vtnerd: As a follow-up I read the above comment, but it's not making any sense. The existing code should be pulling in the same path ... ?

vdo:
FYI I could build for ARM64 arch using this PR, see my branch: 
https://github.com/vdo/monero-lws/commit/c2f6db90c7bb0ff69484636257064eed445ee2b9

Using this Dockerfile on top of `vtnerd/monero-lws:development` I get the build errors @j-berman mentions above, so the changes to CMakeLists.txt are needed, I think.

vtnerd:
@vdo, The problem might be that I typically build on Gentoo against libraries that have been built with `USE` flags that strip out unnecessary dependencies. So I will take a quick scan, and likely just merge.

sethforprivacy:
@vtnerd I just tested this against the latest MyMonero and it is working flawlessly. Anything else needed before we can get this merged? I'd like to setup automatic build/push via Github CI once it's merged in so we can pull images from Docker Hub/GHCR 😄

sethforprivacy (approved):

vtnerd:
@sethforprivacy the problem is this PR is targeted for the 0.17 branch, so it doesn't make any sense to do that. It's probably easiest to merge this now, then do some sort of cherry-pick into master/develop with a change to 0.18, then do an immediate branch for 0.18. Although I'm not really sure how to leave the master/develop branches with this patch because it's better suited _only_ in a release branch.

sethforprivacy:
That's fine, it's a simple 2-line PR to bump to v0.18.0.0 in the Dockerfile once monero-lws as a whole is ready for v0.18.0.0.

vtnerd (approved):

vtnerd:
@sethforprivacy there's a [new branch](https://github.com/vtnerd/monero-lws/tree/release-v0.2_0.18) with the relevant `Dockerfile` changes. The `master` and `develop` branches still reference 0.17 - I guess we should just periodically re-sync those to the hashes on the `monero` `master` branch ...

vdo:
I'm getting errors when building that branch:
https://paste.sethforprivacy.com/?632a27d7e6e7208d#GzoXnKzpwcQrdvnwRk2ir2wKPmcwtZhHvnpRpsWgou7U

vtnerd:
@vdo I think a new issue should be opened for this. It looks like some issue with linking libunbound statically. Presumably other instances used a shared library, which picked up the relevant dependencies of libunbound.

selsta:
You have to compile static unbound manually, like this: https://github.com/sethforprivacy/simple-monerod-docker/pull/61/files

vtnerd:
@vdo does `monerod` even compile with that libunbound? I just tried this on Gentoo and it's complaining about the same missing functions. Gentoo does not expose a `libevent=no` `USE` flag so I can't verify completely, but everything is suggesting this is the case.  This may need fixing upstream if a more robust cmake for statically built libunbound is desired.

vtnerd:
In summary, use a `libunbound.so` or build a custom unbound like @selsta mentioned for now. Otherwise just file an issue, and this may have to get bumped upstream to `monero` as well.

selsta:
@vtnerd We did have an issue open but it was closed: https://github.com/monero-project/monero/issues/8439#issuecomment-1186548894

Do you have an idea of how to fix this upstream properly?

vtnerd:
@vdo @selsta Something like:

```
pkg-config --libs --static --pure libunbound
```
should do the trick on most platforms (assuming unbound is installed). On my Gentoo box, which has nearly everything stripped outputs:
```
-lunbound -lcrypto -lssl -lcrypto -levent
```
Howard probably dropped because adding this to the cmake file might be a different source of pain when building (because `pkg-config` must be installed, etc). Maybe I could do a "test" to see if `pkg-config` is available and only use it on that scenario.

vtnerd:
Scratch all that above, that is pulling information from the dynamic library. The `.pc` file has to be installed somewhere, and it doesn't appear to be (on Gentoo anyway). Gross, static library linking is indeed a pain, which is why Howard basically punted.

vdo:
@vtnerd  Just to be clear, this happens when building using the Dockerfile. I will open a new issue for clarity.

vtnerd:
See PR #39

PR #31, opened 2022-02-21
per_kb_fee -> per_byte_fee

j-berman:
Ok did the easy thing and converted `per_kb_fee` to `per_byte_fee`

One thing: upon deeper inspection, it *looks* like the server shouldn't be masking the fee. Maybe you see something I don't.. but if the client "re-masks" *after* factoring in weight (as MyMonero does [here](https://github.com/mymonero/mymonero-core-cpp/blob/a53e57f2a376b05bb0f4d851713321c749e5d8d9/src/monero_fee_utils.cpp#L233), here's [the caller](https://github.com/mymonero/mymonero-core-cpp/blob/a53e57f2a376b05bb0f4d851713321c749e5d8d9/src/monero_transfer_utils.cpp#L435)), the fee calculation is different than it would otherwise be. Since wallet2 [seems to only mask once after weight is factored in](https://github.com/monero-project/monero/blob/9aab19f349433687c7aaf2c1cbc5751e5912c0aa/src/wallet/wallet2.cpp#L336), and I figure the intention is to match wallet2, then makes sense to let the client do the masking so as to not lead to fingerprinting?

Running the numbers:

```.cpp
uint64_t calculate_fee_from_weight(uint64_t base_fee, uint64_t weight, uint64_t fee_multiplier, uint64_t fee_quantization_mask)
{
  // monero-lws expected calculated fee
  const std::uint64_t per_byte_fee_masked =
    ((base_fee + (fee_quantization_mask - 1)) / fee_quantization_mask) * fee_quantization_mask;
  uint64_t monero_lws_calculated_fee =  weight * per_byte_fee_masked * fee_multiplier;
  monero_lws_calculated_fee = (monero_lws_calculated_fee + fee_quantization_mask - 1) / fee_quantization_mask * fee_quantization_mask;

  // wallet2 calculated fee
  uint64_t fee = weight * base_fee * fee_multiplier;
  fee = (fee + fee_quantization_mask - 1) / fee_quantization_mask * fee_quantization_mask;

  printf("monero-lws fee: %lu \nwallet2 fee: %lu", monero_lws_calculated_fee, fee);
  return fee;
}
```

Outputs:

```
monero-lws fee: 344640000
wallet2 fee:   336650000
```

### DIFF
--- a/src/rest_server.cpp
+++ b/src/rest_server.cpp
@@ -537,15 +537,15 @@ namespace lws
         if (!resp)
           return resp.error();
 
-        if (resp->size_scale == 0 || 1024 < resp->size_scale || resp->fee_mask == 0)
+        if ((resp->size_scale != 1 && resp->size_scale != 1024) || resp->fee_mask == 0)
           return {lws::error::bad_daemon_response};
 
-        const std::uint64_t per_byte_fee =
-          resp->estimated_base_fee / resp->size_scale;
-        const std::uint64_t per_byte_fee_masked =
-          ((per_byte_fee + (resp->fee_mask - 1)) / resp->fee_mask) * resp->fee_mask;
+        const bool use_per_byte_fee = resp->size_scale == 1;
+        const std::uint64_t base_fee = resp->estimated_base_fee;
+        const std::uint64_t base_fee_masked =
+          ((base_fee + (resp->fee_mask - 1)) / resp->fee_mask) * resp->fee_mask;
 
-        return response{per_byte_fee_masked, resp->fee_mask, rpc::safe_uint64(received), std::move(unspent), std::move(req.creds.key)};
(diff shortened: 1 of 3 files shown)


j-berman:
Confirmed that the server shouldn't be masking to maintain parity with wallet2. I confirmed by getting a client-side + server-side implementation to match wallet2 using MyMonero's code. Will be submitting a separate PR to MyMonero's code after a bit more testing

vtnerd:
So the API schema needs to be updated, correct? I just went through some mymonero code and it looks like `per_byte_fee` is now preferred.

j-berman:
latest commit look good? think that's what you meant, would make sense to me

j-berman:
Also will squash before merge. Pretty unnecessary set of commits to keep around

PR #40, opened 2022-07-28
CMakeLists: Remove manual pull-in of already found libraries

j-berman:
Bringing this commit by @TheCharlatan over from #30 also :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,19 +88,36 @@ if (NOT MONERO_BUILD_DIR)
   message(FATAL_ERROR "The argument -DMONERO_BUILD_DIR must specify a location of an existing monero build")
 endif()
 
+if(STATIC)
+  if(MSVC)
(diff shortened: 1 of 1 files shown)


vtnerd:
How much of a pain is it to get this mergeable with the `develop` branch? Arguably this one in particular (which fixes no bugs technically) should be made against that branch first before being pulled into other branches. One difficulty is that I have no CI system to verify this at the moment.

j-berman:
Not a pain :)

But adding, this PR is necessary for the release branch too otherwise the docker build errors out with a missing `libzmq.so` error at runtime.

I think if we squash commits before merging PR's it would be easier to track/cherry-pick all the right stuff going forward

vtnerd:
> I think if we squash commits before merging PR's it would be easier to track/cherry-pick all the right stuff going forward

Agreed. I should've asked for a squashed commit before merging; I lost track of one commit and it caused some havoc.

PR #41, opened 2022-07-28
update develop branch to v0.18

j-berman:
Commit 3aa0a9ec6b7043b0de1c50b404d09fd7404cdab2 in develop needs to use monero v0.18, otherwise the develop branch doesn't compile

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,19 +88,36 @@ if (NOT MONERO_BUILD_DIR)
   message(FATAL_ERROR "The argument -DMONERO_BUILD_DIR must specify a location of an existing monero build")
 endif()
 
+if(STATIC)
+  if(MSVC)
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .lib .dll.a .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+  else()
+    set(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+  endif()
+endif()
+
 load_cache(${MONERO_BUILD_DIR} READ_WITH_PREFIX monero_
   Boost_THREAD_LIBRARY_RELEASE
   CMAKE_CXX_COMPILER
   EXTRA_LIBRARIES
+  LIBUDEV_LIBRARY
+  usb_LIBRARY
   HIDAPI_INCLUDE_DIR
+  HIDAPI_LIBRARY
   LMDB_INCLUDE
   monero_SOURCE_DIR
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -274,17 +222,17 @@ set_property(TARGET monero::libraries PROPERTY
     ${Boost_SERIALIZATION_LIBRARY}
     ${Boost_SYSTEM_LIBRARY}
     ${Boost_THREAD_LIBRARY}
-    ${HIDAPI_LIBRARIES}
-    ${OPENSSL_LIBRARIES}
+    ${monero_HIDAPI_LIBRARY}
+    ${monero_usb_LIBRARY}
+    ${monero_LIBUDEV_LIBRARY}
+    ${ZMQ_LIB}
```
vtnerd, line 228: Is this needed here? It might be needed in `monerod` 0.18+ now, but previously I got away with only linking against ZeroMQ in the binaries that used it (`monero-lws-admin` technically shouldn't need it, I think).

CMakeLists.txt
```diff
@@ -247,17 +197,15 @@ foreach (LIB ${MONERO_LIBRARIES})
...
 
 add_library(monero::libraries INTERFACE IMPORTED)
 set_property(TARGET monero::libraries PROPERTY
   INTERFACE_INCLUDE_DIRECTORIES
     ${Boost_INCLUDE_DIR}
-    ${HIDAPI_INCLUDE_DIRS}
-    ${OPENSSL_INCLUDE_DIR}
-    ${ZMQ_INCLUDE_PATH}
+    ${monero_HIDAPI_INCLUDE_DIRS}
+    ${monero_OPENSSL_INCLUDE_PATH}
```
vtnerd, line 208: Was `ZMQ_INCLUDE_PATH` intentionally dropped? If the library is in a non-standard location, it should fail now.

vtnerd (commented):
Going to merge this one first to do some local testing...

CMakeLists.txt
vtnerd: Nevermind, I see, this is used in the one location where its needed ...

CMakeLists.txt
j-berman: Good spot, builds and runs fine without it. Guessing it was missed how it's already set for `monero-lws-daemon` in `src/CmakeLists.txt`. Removed

vtnerd:
Can you squash these? Github now has a feature where it allows you to compare changes since your last review, and I think it still works with squashing.

vtnerd:
Or no wait, maybe its better with this history, to keep @TheCharlatan in the history.

PR #43, opened 2022-07-29
add arm64 arch support

vdo:
This commit adds support for multi-arch builds, and also a couple of required libs to run.


vtnerd:
This should be done onto the `develop` branch first. The develop branch should now work properly with Docker as well (it is slightly ahead of the 0.18 branch actually).

vtnerd:
And to clarify, I'm requesting a PR onto `develop` instead of `release-v0.2_0.18`.

PR #44, opened 2022-07-29
Detect missing HIDAPI in Cmake

vtnerd:
`HIDAPI` is not needed to build upstream `monerod`. This detects the missing library, and omits it from the include and linking stages. The `Dockerfile` was also updated to omit the installation of `HIDAPI` since its not needed.

I manually verified that this still works when `HIDAPI` is present.


j-berman (approved):
Tested both non-docker and docker build compile and run fine for me

PR #45, opened 2022-07-30
add arm64 arch support

vdo:


Dockerfile
```diff
@@ -114,6 +119,8 @@ RUN apt-get install --no-install-recommends -y \
     ca-certificates \
     curl \
     jq \
+    libhidapi-dev \
```
vtnerd, line 122: These are listed above (and another PR removes `libhidapi-dev` entirely) ... so I think they can be removed?

vtnerd (changes requested):

PR #47, opened 2022-08-18
Remove source and build folder match

LeoNero:
This is a suggestion.

I made this PR because, for example, I have the monero code in a certain folder, and I compile the monero code in a RAM Disk mounted in a separate folder.

Another possible fix is to add a flag to disable this `if` in cases like mine.

If this PR does not make sense, lmk.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,10 +103,6 @@ load_cache(${MONERO_BUILD_DIR} READ_WITH_PREFIX monero_
   ZMQ_LIB
 )
 
-if (NOT (monero_monero_SOURCE_DIR MATCHES "${MONERO_SOURCE_DIR}(/src/cryptonote_protocol)"))
-  message(FATAL_ERROR "Invalid Monero source dir - does not appear to match source used for build directory")
-endif()
-
 if (NOT (CMAKE_CXX_COMPILER STREQUAL monero_CMAKE_CXX_COMPILER))
   message(FATAL_ERROR "Compiler for monero build differs from this project")
 endif()


trasherdk:
I'm building both `monero` and `monero-lws` out-of-source, with no problems.

```sh
cmake \
-DMONERO_SOURCE_DIR="$(realpath "${STAGE}/../monero/monero-src")" \
-DMONERO_BUILD_DIR="$(realpath "${STAGE}/../monero/monero-build")" "${SOURCE}" \
&& { echo "${WHITE}* CMake finished${GREEN} OK ${RESTORE}"; } \
|| { echo "${WHITE}* CMake finished${RED} FAIL ${RESTORE}"; exit 1; }

make -j$(nproc)
```

LeoNero:
It does not work for me. My source dir is `/Users/me/Documents/monero/monero/` and my build dir is `/Volumes/MoneroRAMDisk/build-monero-clang/`. I think it maybe works for you because both are inside `monero/`?

I am not used to CMAKE and haven't had much time to understand all the `CMakeLists.txt` yet.

trasherdk:
@LeoNero Nope, that's not how stuff works. It's just variables being assigned a value.
It does require the paths to exist and contain the expected files.

vtnerd:
That check ensures that the monero build directory used the sane monero source directory for the build. In other words, it is checking (to prevent) multiple monero source directories. Out of source builds are definitely possible, I do it all the time to test locally.

LeoNero:
I see. It makes total sense, thanks! I am closing this PR then, and will continue to just remove the check manually locally. In case this is a issue to someone else, then this PR can be reopened and discussed later.

PR #52, opened 2022-09-27
Fixed integer conversion in wire::

vtnerd:
@jeffro256 found a bug in the integer conversion code for wire. The code was hand-rolled and therefore suspect. The intent was to reduce the ASM noise per integer type, so that inlining was more likely.

I have switched to `boost::conversion::try_lexical_convert` which produces ASM compaction comparable to what this incorrect version had.

I'm not aware of any security issues, although technically this is UB here. The biggest fallout is that incorrect integer values will not be properly reported, but each value requires further checking typically anyway.

If someone approves this, I will just cherry-pick onto `master` and the current release branch.

### DIFF
--- a/src/wire/read.cpp
+++ b/src/wire/read.cpp
@@ -35,9 +35,12 @@ void wire::reader::increment_depth()
     WIRE_DLOG_THROW_(error::schema::maximum_depth);
 }
 
-[[noreturn]] void wire::integer::throw_exception(std::intmax_t source, std::intmax_t min)
+[[noreturn]] void wire::integer::throw_exception(std::intmax_t source, std::intmax_t min, std::uintmax_t max)
 {
-  WIRE_DLOG_THROW(error::schema::larger_integer, source << " given when " << min << " is minimum permitted");
+  if (source < 0)
+    WIRE_DLOG_THROW(error::schema::larger_integer, source << " given when " << min << " is minimum permitted");
+  else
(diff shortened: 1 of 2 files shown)


jeffro256:
` try_lexical_cast` returns true even when converting a negative value in a signed int to an unsigned int

vtnerd:
I knew this at some point about `lexical_cast` and forgot too - I recall it biting another developer because its non-obvious that conversions between integer types would have `stringstream` behavior, but that's why `boost::numeric` exists I guess.

vtnerd:
All we need is this right here:
```c++
template<typename To, typename From>
bool WillOverflow(From val) {
  assert(std::numeric_limits<From>::is_integer);
  assert(std::numeric_limits<To>::is_integer);
  if (std::numeric_limits<To>::is_signed) {
    return (!std::numeric_limits<From>::is_signed &&
              (uintmax_t)val > (uintmax_t)INTMAX_MAX) ||
           (intmax_t)val < (intmax_t)std::numeric_limits<To>::min() ||
           (intmax_t)val > (intmax_t)std::numeric_limits<To>::max();
  } else {
    return val < 0 ||
           (uintmax_t)val > (uintmax_t)std::numeric_limits<To>::max();
  }
}
```

[lol @ c++ integer rules](https://blog.reverberate.org/2012/12/testing-for-integer-overflow-in-c-and-c.html)

PR #53, opened 2022-10-08
Added unit tests, and fixed two bugs:

vtnerd:
* Integer conversion checks in src/wire/read.h
 * Missing "boolean" function in wire::writer and derived types

This is similar to #52 but should fix the integer checking issue, and adds unit tests for the wire code. Eventually the two fixes will be PR'ed onto the release branch, but the unit tests will likely not (at least not initially).

I doubt many people will want to go through this review, so I will leave it here for ~2 weeks and then just merge. Look out for `lest.hpp` which is from another project, but somewhat easy to review.

### DIFF
--- a/src/wire/read.h
+++ b/src/wire/read.h
@@ -83,7 +83,7 @@ namespace wire
     //! \throw wire::exception if next value not a boolean.
     virtual bool boolean() = 0;
 
-    //! \throw wire::expception if next value not an integer.
+    //! \throw wire::exception if next value not an integer.
     virtual std::intmax_t integer() = 0;
 
     //! \throw wire::exception if next value not an unsigned integer.
@@ -104,8 +104,7 @@ namespace wire
     //! \throw wire::exception if next value invalid enum. \return Index in `enums`.
     virtual std::size_t enumeration(epee::span<char const* const> enums) = 0;
 
-    /*! \throw wire::exception if next value not array
-        \return Number of values to read before calling `is_array_end()`. */
+    //! \throw wire::exception if next value not array
     virtual std::size_t start_array() = 0;
 
     //! \return True if there is another element to read.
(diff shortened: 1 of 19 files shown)


jeffro256:
The bug fixes look good, but I don't know anything about `lest`. Is there an upstream repo you could add as a submodule?

vtnerd:
There is a [github repo](https://github.com/martinmoene/lest), but I'm not sure what a submodule provides in this scenario given that just a single file that is needed.

vtnerd:
BTW [compare the changes here](https://github.com/vtnerd/monero-lws/compare/b40b67d5b10cf285901808ec0ca53f42c40ac0de..0c3eef290c3e3121d69b0d1b41c211b74882639d).

vtnerd:
Also - I might make the type templated again after a bit of thought. This will help with performance optimizations, etc.

serhack (approved):
It looks fine for me. I do not have skills to comprehend deeply the changes, but at my eyes it seems fine :)

vtnerd:
Moved back to templated source for the casting functions.

jeffro256:
I don't know if you should move towards splitting the conversion function into different types of functions. The whole reason I found the bug is that the new JSON code in the `wire` module in the `monero` project used an integer conversion going from `int64_t` to `uint64_t` (IIRC). At some point you may need a `cast_unsigned_to_signed` and `cast_signed_to_unsigned`? I liked the `convert_to` function as it was... maybe I could reach out to the blogger and ask for copyright clarification?

jeffro256:
Sorry for the delay of this comment o_0. Are you planning to use this integer conversion code for the `monero` serialization module?

vtnerd:
Yes.

jeffro256:
K then I guess I'll just check how that code uses these integer conversion functions when I get there.

src/wire/read.h
```diff
@@ -167,76 +166,58 @@ namespace wire
...
+      using limit = std::numeric_limits<T>;
+      static_assert(
+        std::is_signed<T>::value && std::is_integral<T>::value,
+        "target must be signed integer type"
+      );
+      static_assert(
+        std::is_signed<U>::value && std::is_integral<U>::value,
+        "source must be signed integer type"
+      );
+      if (source < limit::min() || limit::max() < source)
```
jeffro256, line 184: This is not guaranteed to work if you are casting from a large int to small int

src/wire/read.h
```diff
@@ -167,76 +166,58 @@ namespace wire
...
+      using limit = std::numeric_limits<T>;
+      static_assert(
+        std::is_unsigned<T>::value && std::is_integral<T>::value,
+        "target must be unsigned integer type"
+      );
+      static_assert(
+        std::is_unsigned<U>::value && std::is_integral<U>::value,
+        "source must be unsigned integer type"
+      );
+      if (limit::max() < source)
```
jeffro256, line 201: Same here: This is not guaranteed to work if you are casting from a large int to small int

jeffro256 (changes requested):
When casting ints, you need to cast it into an intermediate type which is guaranteed to be able to represent both `T` and `U`, you can't only use `T` limits, because `U` might be bigger than those limits, in which case using the `<` and `>` operators is UB.

src/wire/read.h
vtnerd: There are unit tests for this, so you will need to provide a quick example of this not working. You seem to be claiming that `source` will get downpromoted to the other operand; the `limit::min()` and `limit::max()` value should get promoted to the larger size, per C/C++ rules. 

A quick grab from [cpp-reference](https://en.cppreference.com/w/c/language/conversion) on implicit conversions in arithmetic operations (finding the "common type"):

> If the types have the same signedness (both signed or both unsigned), the operand whose type has the lesser conversion rank1 is implicitly converted2 to the other type.

Typically the issue between casting is between signed and unsigned values, checks of the same signedness are typically easier.

PR #54, opened 2022-10-08
Fixed two bugs in wire:

vtnerd:
* Integer conversion checks in src/wire/read.h
  * Missing "boolean" function in wire::writer and derived types

This is the same as #53 but without the unit tests and onto the 0.2 release branch. Trying to strip the review to its core components to make it easier.

### DIFF
--- a/src/wire/json/write.cpp
+++ b/src/wire/json/write.cpp
@@ -73,6 +73,12 @@ namespace wire
     return buf;
   }
 
+  void json_writer::boolean(const bool source)
+  {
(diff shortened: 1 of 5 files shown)


vtnerd:
Also, the `will_overflow` is not `constexpr` on this branch because the `CXX14` patch was not merged.

vtnerd:
Force pushed a timestamp correction on the commit, and nothing else.

jeffro256 (approved):
Looks good!

PR #60, opened 2022-12-05
Adding vtnerd GPG signing key to repo

vtnerd:
This public key was just used to sign 0.2.0 last night.


selsta:
Maybe extras? https://api.csswg.org/bikeshed/?force=1&url=https://raw.githubusercontent.com/vector-of-bool/pitchfork/develop/data/spec.bs#tld.extras

vtnerd:
I'm leaning towards `misc` due to the link provided by @selsta . The `extras` directory is supposed to be submodules, and I don't see that happening with `gpg_keys` (for a little while).

In other words, I'll update this PR to make it `misc/gpg_keys/vtnerd.asc` and perhaps in the future we'll have a top-level repo in `monero-project/gpg_keys` at which point this would be `extras/gpg_keys/vtnerd.asc`.

vtnerd:
FYI - I'm trying to go with the pitchfork layout because I like what its trying to do, etc.

PR #63, opened 2023-02-14
Adding msgpack support to ::wire:: library

vtnerd:
Support for webhooks is in the works, which requires variable sized records stored in the DB. I have decided to store this data using msgpack encodings.

This encoding library supports both integer and string keys for object types. The reader can transparently support both, while the writer requires a flag to toggle integer or string key output.

Types must specify an integer key to use with msgpack, so some existing code wouldn't work correctly with msgpack directly. The webhook DB records will use integer keys for performance/compaction reasons.


vtnerd:
Fixed Linux build with force-push.

vtnerd:
Im doubting any reviews of this, to just merging :/

PR #64, opened 2023-03-25
Decoy selection algo off-by-1

j-berman:
[wallet2](https://github.com/monero-project/monero/blob/f307621678c24ee7ca158fde38080d787d678733/src/wallet/wallet2.cpp#L6454-L6455) + [consensus](https://github.com/monero-project/monero/blob/f307621678c24ee7ca158fde38080d787d678733/src/cryptonote_core/blockchain.cpp#L3531-L3536) allow a user to spend an output that's in a block 9 blocks prior to the final block in the chain, since the output is eligible to be included in the *next* block at that point (see discussion [here](https://github.com/monero-project/monero/pull/8794#issuecomment-1478116268)). The decoy selection algo should therefore apply the same criteria when selecting decoys.

The PR to change the decoy selection algo in the monero repo is still open here: https://github.com/monero-project/monero/pull/8794

### DIFF
--- a/src/util/gamma_picker.cpp
+++ b/src/util/gamma_picker.cpp
@@ -68,7 +68,8 @@ namespace lws
 
   bool gamma_picker::is_valid() const noexcept
   {
-    return CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE < rct_offsets.size() + 1;
+    static_assert(CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE > 0);
+    return CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE - 1 < rct_offsets.size();
   }
 
   std::uint64_t gamma_picker::spendable_upper_bound() const noexcept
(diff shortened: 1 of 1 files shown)


src/util/gamma_picker.cpp
```diff
@@ -85,7 +91,7 @@ namespace lws
 
     static_assert(std::is_empty<crypto::random_device>(), "random_device is no longer cheap to construct");
     static constexpr const crypto::random_device engine{};
-    const auto end = offsets().end() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE;
+    const auto end = offsets().end() - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE + 1;
```
vtnerd, line 94: Add a `static_assert` such that `CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE > 0`.

src/util/gamma_picker.cpp
```diff
@@ -68,14 +68,20 @@ namespace lws
 
   bool gamma_picker::is_valid() const noexcept
   {
-    return CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE < rct_offsets.size();
+    return CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE < rct_offsets.size() + 1;
```
vtnerd, line 71: I would `-1` on  other side of equation, with a `static_assert` here (such that macro constant must be greater than 0.

vtnerd (changes requested):
A couple of small change requests for robustness.

vtnerd:
This was cherry picked onto the `master` and `release-v0.2` branches.

PR #67, opened 2023-04-20
Fix Slackware build

trasherdk:
Fix Slackware build - Add missing `#include <limits>`


trasherdk:
Woops, wrong branch. Sorry.

PR #68, opened 2023-04-20
Fix Slackware build

trasherdk:
- Add missing `#include <limits>`

### DIFF
--- a/src/wire/msgpack/base.h
+++ b/src/wire/msgpack/base.h
@@ -29,6 +29,7 @@
 
 #include <cstdint>
 #include <string>
+#include <limits>
 #include <tuple>
 
 #include "byte_slice.h"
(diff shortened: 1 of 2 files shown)


vtnerd:
Sorry to nitpick, but I believe this is [where the include should be](https://github.com/vtnerd/monero-lws/blob/develop/src/wire/msgpack/base.h#L31). Not sure why it works out this way. 

Could you retry with the other file? I don't have a slackware machine currently.

trasherdk:
Done.

I just slapped the include in where the build complained about missing it.

While we are nitpicking, cmake's feeling get hurt, when relative paths are used in `MONERO_SOURCE_DIR` and `MONERO_BUILD_DIR`. The `~/` might be an exception for some reason.

What works on `Slackware`:
```sh
monero-lws/build$ cmake -DMONERO_SOURCE_DIR="$(realpath ../../monero)" -DMONERO_BUILD_DIR="$(realpath ../../monero/build)" ..
```

PR #72, opened 2023-06-03
Add zero-confirmation support to webhooks (only)

vtnerd:
Getting zero-confirmation notifications via webhooks requires the `--sub` option (and `monerod` must have `--zmq-pub` set as well). The transactions _do not_ show up in the REST API until 1 confirmation. The notifications are nearly instant (everything uses pub/sub instead of polling).

Zero-confirmations in the REST API _might be_ added in the future but that is a little more involved (unless I can come up with an alternative implementation).

The block number and output numbers are omitted from zero-conf output, since they lack valid values.

### DIFF
--- a/src/db/data.cpp
+++ b/src/db/data.cpp
@@ -34,6 +34,7 @@
 #include "wire/json/write.h"
 #include "wire/msgpack.h"
 #include "wire/uuid.h"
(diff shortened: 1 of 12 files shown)


vtnerd:
I did a rebase on the branch - the changes to `::wire::` needed to be included in the PR.

PR #79, opened 2023-07-24
Webhooks for New Accounts

vtnerd:
Adds webhooks on new account creation _request_. Currently the server must be polled for new account creation requests. Auto-accepting all account creation requests still not possible (yet!).


vtnerd:
Rebased after the ZMQ-PUB merge. New accounts creation requests are now notified over optional ZMQ PUB/SUB socket.

PR #80, opened 2023-08-09
Add RabbitMQ support for payment hooks

vtnerd:
This will need a rebase after #79 gets merged. RabbitMQ is an optional dependency that defaults to "OFF".

The CI has **not** been updated yet to verify RabbitMQ enabled builds (it only checks the OFF case).


vtnerd:
Performed a rebase after the new account webhook PR was merged.

PR #86, opened 2023-11-20
Add option to auto-accept new account creation

vtnerd:
The title says it all - you no longer need glue code to accept new addresses/accounts. The downside is there is **no limit** on number of accounts, but there is a limit on the number of pending creation requests. 

Also note that I have yet to implement an account deletion function! So I would only use this option with extreme caution, as someone could fill-up your database (account deletion is a feature to be worked on shortly).


vtnerd:
Rebased, and awaiting CI system before merging.

PR #92, opened 2024-01-23
Refuse chain rollback past a checkpoint

vtnerd:
This prevents the daemon from attempting to fork the blockchain past a hard-coded checkpoint. The daemon cannot rollback past a checkpoint, so this can only happen with a "bad" daemon.

### DIFF
--- a/src/db/storage.cpp
+++ b/src/db/storage.cpp
@@ -1453,7 +1453,7 @@ namespace db
 
         if (*hash != chain.front())
         {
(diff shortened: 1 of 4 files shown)


vtnerd:
The updated check is only done where a chain split is detected. The split must occur after the last checkpoint, or an error is returned. This is not configurable, and only prevents a large invalid reorg; a daemon can still provide an invalid chain after the last checkpoint.

PR #93, opened 2024-02-06
Basic "chain hardening" for slightly untrusted daemons

vtnerd:
This adds basic chain hardening to the daemon, for slightly untrusted daemons. It is still recommended that you run your own node, as there are more features like JSON memory unpacking limits (similar to what the daemon had to endure).

When `--untrusted-daemon` is enabled:
 * Every block `prev_id` is verified, to ensure a chain
 * Every "hard-coded" checkpoint has its block hash verified
    * The two above mean that you cannot easily spoof old blocks - only blocks newer than a checkpoint
 * Every transaction hash is computed and verified against the block hash list
 * The PoW is verified for blocks newer than any checkpoints, forcing a malicious daemon to spend resources to generate fake blocks
 * The block height, hash, and difficulty are periodically printed to help assist in detecting malicious daemons

### DIFF
--- a/src/db/data.cpp
+++ b/src/db/data.cpp
@@ -204,6 +204,43 @@ namespace db
   namespace
   {
     template<typename F, typename T>
+    void map_block_difficulty(F& format, T& self)
+    {
+      wire::object(format, WIRE_FIELD_ID(0, high), WIRE_FIELD_ID(1, low));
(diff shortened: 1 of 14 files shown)


vtnerd:
Force pushed a rebase.

PR #97, opened 2024-02-23
ZMQ pub blocks

vtnerd:
A little different than I intended, but this adds a ZMQ-PUB for recently scanned addresses. The address is sent along with the last block height and hash. This is useful for tracking the internal scan state in real-time, as each thread reports independent results.


vtnerd:
Rebased so that CI can cycle through with updated CMake.

PR #98, opened 2024-02-25
Update chain rollback log message

vtnerd:
I found an error case with the invalid chain rollback detection - when the daemon is still syncing it sends older blocks. The easiest thing to do (for now) is to update the error message to let the user know they must sync the daemon before using LWS.


vtnerd:
Rebased so that CI can cycle through with updated CMake.

PR #101, opened 2024-02-28
ZMQ Pub Spends

vtnerd:
This adds support to webhook/zmq for spend notification. The use case is detecting unexpected spending of funds.


vtnerd:
I just realized this needs some work to be useful. The `tx_pub` from the original output needs to be sent along with the information, to verify whether it was spent.

vtnerd:
Changes:
  * Added source (output) information to zmq pubs
  * Added documentation
  * Fixed `webhook_add`

vtnerd:
Rebased after recent merges.

PR #114, opened 2024-04-23
Fix divide by zero in subaddresses n_maj

vtnerd:
Found by @everoddandeven

### DIFF
--- a/src/rest_server.cpp
+++ b/src/rest_server.cpp
@@ -713,16 +713,16 @@ namespace lws
         const std::uint32_t n_major = req.n_maj.value_or(50);
         const std::uint32_t n_minor = req.n_min.value_or(500);
         const bool get_all = req.get_all.value_or(true);
-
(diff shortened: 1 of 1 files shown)


everoddandeven:
signal-2024-04-26-181317_003
signal-2024-04-26-181317_002
signal-2024-04-26-181146_003
signal-2024-04-26-181146_002
I think I have founde other bugs. Now all subaddrs methods are crashing server with `2024-04-26 16:16:13.995	D Schema expected array: 2 array elements of at least 2 bytes each exceeds 3 remaining bytes
terminate called after throwing an instance of 'std::system_error'
  what():  thrown at storage.cpp:1014: Schema expected array
` error

vtnerd:
Could you provide some steps to reproduce this behavior? I just ran both of the `/provision_subaddrs` requests and it worked as expected.

I have a feeling that reproducing the DB state might be difficult.

PR #118, opened 2024-05-29
Add support for remote scanning via custom TCP

vtnerd:
This is the first cut at remote scanning in LWS. It took a while to complete because a custom TCP implementation using ASIO was done - ZMQ sockets were too limited in their capabilities for the task.

There's lots of code here, so I will leave for a bit as I perform additional testing on this, including performance tests. I'm still uncertain on how this will perform with lots of accounts, etc.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,7 @@ target_link_libraries(monero-lws-daemon-common
     monero-lws-common
     monero-lws-db
     monero-lws-rpc
+    monero-lws-rpc-scanner
     monero-lws-wire-json
     monero-lws-util
     ${Boost_CHRONO_LIBRARY}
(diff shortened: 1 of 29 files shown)


vtnerd:
This last force push:
   * Rebased against `develop`
   * Fixed exception bug for `socket.remote_endpoint` calls
   * Fixed exception bug for `socket.close` calls
   * Called `socket.shutdown` before `socket.close` for better platform compliance
   * Use libsodium password hashing library and memory compare to prevent password leakage.

vtnerd:
Finally ran a timed comparison test with 1000 accounts. A Ryzen 3900x saw a 5 minute drop in scan time when two laptops were added to the mix. The laptops took longer, but that is only due to the comparatively slower single threaded CPU performance. The added stress on `momerod` didn't impact performance negatively. 

Looks like this patch will be merged, as it works as expected. I may tweak some of the parameters - most notably the threshold for remote scanning to be initiated.

PR #119, opened 2024-06-03
Fix several bugs:

vtnerd:
* lws::account height update should only go up.
  * Webhook confirmations can start after first new block
  * Webhook confirmations could face a rescan

Found these issues while digging deeper in the remote scanning code (although not related to that code). This will get merged relatively soon as these bugs need to be back-ported to release branch.


vtnerd:
Added some more tests, and fixed another issue within the DB scan height (it previously wasn't an issue until changes in the `lws::account` object).

PR #120, opened 2024-06-06
Fix scan height (db) bug on account check-ins

vtnerd:
The last PR still had another bug involving a scan height index. I added unit tests to ensure this index is correctly update (or not changed) during complex scanning conditions (rescans, forks, duplicate check-ins).

### DIFF
--- a/src/db/storage.cpp
+++ b/src/db/storage.cpp
@@ -859,6 +859,41 @@ namespace db
     return out;
   }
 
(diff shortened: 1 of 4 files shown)


vtnerd:
Also note that this fixes some cases where duplicate webhooks could be triggered. The unit tests verify that the web hook only invokes once, unless a rescan is requested (which resets the tracking for webhooks).

vtnerd:
Merged this sooner rather than later, as the bug fix needs to be back ported to release branch.

PR #124, opened 2024-06-24
Add (non-standard) 'daemon_status' endpoint to REST API

vtnerd:
Requested privately. Mostly a relay to from the daemon. Possibly useful to clients that only have access to LWS REST.

### DIFF
--- a/src/rest_server.cpp
+++ b/src/rest_server.cpp
@@ -37,6 +37,7 @@
 
 #include "common/error.h"          // monero/src
 #include "common/expect.h"         // monero/src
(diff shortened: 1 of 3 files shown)


vtnerd:
Force pushed a change where a daemon timeout is **not** reported as a 503 error, and is instead reported in the `state` field of the response. All other endpoints still use 503.

vtnerd:
Force pushed a rebase + fixes for the timeout.

vtnerd:
Force pushed another rebase after recent ZMQ bug fixes.

PR #131, opened 2024-09-10
Add equivalent check to validate admin endpoint

vtnerd:


vtnerd:
Crap the CI isn't useful right now as the upstream macos+`boost::uuid` issue is hitting here. Will just merge-in, and assume it still works.

PR #133, opened 2024-09-20
/login update access_time, status

TheTrunk:
Upon calling /login endpoint update
  - access_time
  - status to active
  
This is a step for automation that I think many users might appreciate. We are used to put accounts in an inactive status after some time period has passed. When user logs in back into the solution, we want their account to become active again. These little changes are thus helpful and needed for our automation.

Ideally this should be behind some flag, just raising the PR as I got other devs asking to show an example solution.

### DIFF
--- a/src/rest_server.cpp
+++ b/src/rest_server.cpp
@@ -701,6 +701,17 @@ namespace lws
             if (is_hidden(account->first))
               return {lws::error::account_not_found};
 
+            // Update access time
+            auto update_result = disk.update_access_time(req.creds.address);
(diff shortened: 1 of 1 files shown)


vtnerd:
Can this be re-opened on the `develop` branch? I'll then backport to `master` and `release-v0.3_0.18` branches.

vtnerd:
Nevermind, I see this was done mainly as an example. I'm not sure if the default behavior should be to do another disk I/O to change state into active. The other call - to update access time - should likely be added upstream.

PR #136, opened 2024-10-03
Switch from epee http server to boost::beast http server. Min boost 1.70

vtnerd:
# Comments in log
There is roughly a 7.4% increase in performance in the switch to boost::beast. Additionally, the REST endpoints `/daemon_status`, `/get_unspent_outs`, and `/submit_raw_tx` do not block in ZMQ calls, allowing for better response times regardless of `monerod` status.

The REST endpoints `/login` and `/get_random_outs` still need updates to prevent blocking (`/login` is conditional on DB state).
 
# Additional Commentary
The REST endpoints `/daemon_status`, `/get_unspent_outs`, `/get_random_outs`, `/get_unspent_outs`, and `/submit_raw_tx` all require calls into `monerod` to complete. The endpoint `/login` _may_ perform blocking HTTP I/O if webhooks are configured for new accounts. Some of the values can be cached, but even then all of the mentioned endpoints could be temporarily blocked by a slow, unresponsive, or unavailable `monerod` (or slow HTTP server for `/login` webhooks). This hurts throughput on the REST threads, especially the endpoints `/get_address_info`, `/get_address_txs`, and (sometimes) `/login` which require disk I/O but no outside information.

Since the default is 1 REST thread, these blocking events can give the appearance of an unresponsive REST server when it is `monerod` (or outside HTTP server) that is the source of the slowdown. This patch (begins) to alleviate the issue by making `/daemon_status`, `/get_unspent_outs`, and `/submit_raw_tx` asynchronous when blocking I/O into `monerod` needs to be performed. While waiting for `monerod` response, additional requests to the same endpoint will queue and then perform a batch response when the `monerod` I/O completes. In the best case scenario, this takes 2-3 milliseconds, but can be longer for various reasons.

Additionally, `/daemon_status` and `/get_unspent_outs` now cache `monerod` responses for 5 seconds to reduce unnecessary burden on `monerod`. This improves the throughput of these endpoints dramatically, and `wrk2` load stressing is now somewhat pointless on `/daemon_status` in particular since it generates a cache response instantly.

# Implementation Details
A REST endpoint can now be: (1) full synchronous (**all** admin endpoints, `/get_address_info`, `get_address_txs`, and `/login` _eventually_), (2) partially synchronous (`/daemon_status`, and `/get_unspent_outs`) or (3) fully asynchronous (`/get_random_outs` _eventually_, and `get_unspent_outs`). These three possibilities help with micro-optimizations internally on performance:
  * Fully synchronous endpoints never allocates memory for `std::function` callback
  * Partially asynchronous sometimes uses the faster templated callback (as opposed to `std::function` callback). Basically, some checks have to be done for synchronous endpoints, and `std::function` is duplicating the checks (in addition to being slightly slower in invocation).
  * Fully asynchronous always uses the type-erased `std::function` callback

This patch doesn't have code comments describing how this works, I will most likely update this PR in the future with such information (before merging). Additionally, `/submit_raw_tx` still needs additional stress testing.

Otherwise, this is a good end-of-month PR as its 99% complete.


vtnerd:
Force pushed some bug fixes. The build will still fail, that will be updated in a separate PR.

vtnerd:
Force pushed a few modifications - changed how the strand wrapping works. Also updated a comment.

vtnerd:
Force pushed a rebase (checks pass!), and 2 bug fixes.

vtnerd:
Force pushed Boost dependencies update.

vtnerd:
Noticed some whitespace differences in `zmq_asnyc.cpp`, corrected.

PR #149, opened 2024-10-27
Update Dockerfile and CMake files to build develop/master

vtnerd:
The `Dockerfile` will build against the `develop` branch once this [upstream PR](https://github.com/monero-project/monero/pull/9530) is merged.


vtnerd:
There is a difference between upstream monero branches and how `libsodium` gets pulled in. So this code now reacts to either, so that the `Docker` build should work as well as `master` builds.

PR #150, opened 2024-10-28
Switch from epee http client to boost::beast. All HTTP now non-blocking.

vtnerd:
As the title suggests, this removes all blocking epee HTTP client code and moves to an async `boost::beast` implementation. The REST endpoint `/login` is now always non-blocking, even with account creation webhooks. The scanner no longer blocks for webhooks, and instead queues up requests. The rate exchange fetch code is also now async.

### DIFF
--- a/src/client_main.cpp
+++ b/src/client_main.cpp
@@ -195,7 +195,7 @@ namespace
   {
     std::shared_ptr<lws::rpc::scanner::client> client_;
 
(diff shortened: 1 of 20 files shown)


vtnerd:
Forgot to include new files. Force pushed a change with them.

vtnerd:
Force push some missing includes fixes.

vtnerd:
Force pushed a rebase after PR merges. Also updated new `io_service` usage to `io_context` after last PR.

PR #166, opened 2025-06-13
Add support for dynamic fees over ZMQ

vtnerd:
As the title says. This requires a corresponding patch in the daemon (on `master`), so the CI should fail. The CI will be re-run once the associated patch gets merged, and then this PR will get merged after.


vtnerd:
Force pushed a dummy change to get the CI to re-run.

vtnerd:
Update the `external/monero` reference, but there should still be another build error.

PR #171, opened 2025-06-22
Set longer timeouts for requests after login

vtnerd:
The change to `boost::beast` changed how timeouts were handled by the server. Each new REST request had a 5-second timeout. This prevented clients from having long running connections. This made things in `lwsf` not work properly, as the connection was terminated quite aggressively server-side.

The new change is to have aggressive 5-second timeouts for **new** connections, and then longer 5-minute timeouts for sessions with a valid login. Neither value is configurable (yet), but can be extended easily to handle that.

### DIFF
--- a/src/rest_server.cpp
+++ b/src/rest_server.cpp
@@ -119,7 +119,7 @@ namespace lws
         sync()
     {}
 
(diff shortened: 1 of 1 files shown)


vtnerd:
Found a bug in the `/login` endpoint, force pushed the change to watch the CI status.

PR #193, opened 2025-11-03
Add regtest option

everoddandeven:
This PR adds a new `regtest` option in order to disable checkpoints check at storage::sync_chain when running a `monerod` in regtest mode, useful in testing environments.

### DIFF
--- a/src/client_main.cpp
+++ b/src/client_main.cpp
@@ -246,7 +246,7 @@ namespace
 
         if (!users.empty())
         {
-          static constexpr const lws::scanner_options opts{false, false};
+          static constexpr const lws::scanner_options opts{false, false, false};
 
           auto new_client = MONERO_UNWRAP(zclient.clone());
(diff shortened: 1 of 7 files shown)


everoddandeven:
> When would this be used?

This would allow us to run lws in (reg)test environments, without having to wait for the fakechain to reach the last checkpoint before we can test lws integration.

https://github.com/btcpay-monero/monero-lws-csharp/pull/3

In that test, the http client fails to connect to lws due to `Resource temporarily unavailable`, and when I run tests locally I see lws complaining about `Attempting rollback past last checkpoint. Wait until daemon finishes syncing - otherwise daemon is performing an attack.`

As you can see we are using monerod in regtest mode:
https://github.com/btcpay-monero/monero-lws-csharp/pull/3/files#diff-c3712128764252a94bd4bc3accae56d3c0564a452ab53f9f835368f9ce790d1e

PR #204, opened 2025-11-17
Return pool transactions in get_address_txs

swansontec:
The basic design is to keep pool transactions in RAM, and then re-scan them on each REST request. The pool is generally pretty small and dynamic, so this strategy is usually cheaper than trying to index everything in the database. Odds are, pool transactions don't belong to any of our addresses, and so they will never even be queried. Thus, any indexing work we do ahead of time is likely to be useless.

I'm submitting this draft PR to show the basic direction this work is heading. There are a number of remaining flaws that need more work:

- I want to add a small cache to the mempool class, which maps from recent addresses to scan results. Repeated wallets requests will hit the cache instead of re-scanning the same transactions over and over.
- While the pool sync mostly works, the tx pool updater will break if the connection drops. I still need to replicate the scanner's restart-on-error logic, and possibly move the update logic into the scanner itself.
- Actually, mempool is the wrong term - maybe I should rename the class to txpool instead.

### DIFF
--- a/src/mempool.cpp
+++ b/src/mempool.cpp
@@ -1,0 +1,215 @@
+// Copyright (c) 2024, The Monero Project
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are
+// permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of
+//    conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list
+//    of conditions and the following disclaimer in the documentation and/or other
+//    materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be
+//    used to endorse or promote products derived from this software without specific
+//    prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+// MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+// THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+// STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+// THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "mempool.h"
+
+#include "common/error.h"
+#include "cryptonote_basic/cryptonote_format_utils.h"
+#include "error.h"
+#include "misc_log_ex.h"
+#include "rpc/daemon_messages.h" // external/monero/src
+#include "rpc/daemon_zmq.h"
+#include "rpc/json.h"
+#include "util/ownership_test.h"
+
+namespace lws
+{
+  const size_t max_cache_size = 1000;
+
+  std::shared_ptr<const mempool::pool_table> mempool::get_txs() const
+  {
+    const std::lock_guard<std::mutex> lock{mutex_};
+
+    // return the existing snapshot if we have one
+    if (snapshot_) return snapshot_;
+
+    // otherwise make a fresh snapshot
+    snapshot_ = std::make_shared<const pool_table>(state_);
+    return snapshot_;
+  }
+
+  void mempool::add_txs(epee::span<cryptonote::transaction> txs)
+  {
+    const std::lock_guard<std::mutex> lock{mutex_};
+    snapshot_.reset();
+
+    for (auto& tx: txs)
+    {
+      auto hash = get_transaction_hash(tx);
+      auto found = state_.find(hash);
+      if (found != state_.end()) continue;
+      state_.insert({
+        hash,
+        std::make_shared<const cryptonote::transaction>(std::move(tx))
+      });
+      // MDEBUG("Mempool added tx " << hash << " size: " << state_.size());
+    }
+  }
+
+  void mempool::reset_txs(pool_table&& txs)
+  {
+    const std::lock_guard<std::mutex> lock{mutex_};
+    snapshot_.reset();
+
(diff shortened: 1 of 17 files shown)


src/mempool.cpp
```diff
@@ -0,0 +1,276 @@
...
+        return;
+      }
+
+      const std::lock_guard<std::mutex> lock{self.mutex_};
+      auto next = std::make_shared<mempool::txs>(*self.current_);
+      next->reserve(parsed->txes.size());
+      for (auto& tx: parsed->txes)
+      {
+        MDEBUG("Mempool tx " << get_transaction_hash(tx) << " size: " << next->size());
+        auto shared_tx = std::make_shared<const cryptonote::transaction>(std::move(tx));
```
vtnerd: Why the `shared_ptr` to the transaction, instead of just having the tx itself in the vector?

src/util/ownership_test.h
```diff
@@ -0,0 +1,81 @@
...
+
+  private:
+    spend_action on_spend;
+    output_action on_output;
+
+    struct detail{
+      db::storage_reader reader;
+      db::cursor::subaddress_indexes cur;
+    };
+    std::optional<detail> subaddress;
```
vtnerd: Switch to `boost::optional` as this might get used by the `release-v0.18` which is only on C++14.

src/mempool.h
```diff
@@ -0,0 +1,83 @@
...
+
+    expect<void> start(rpc::context& context);
+    void stop();
+
+  private:
+    struct thread_detail;
+
+    static crypto::hash get_hash(cryptonote::transaction const& tx);
+
+    mutable std::mutex mutex_;
```
vtnerd: Use `boost::thread` and `boost::mutex` for consistency. We can do a universal change later.

vtnerd (changes requested):
A semi-quick review.

vtnerd (changes requested):
I'm still holding off on some more critical reviews as I think this design through, primarily I'm not sure if scanning on-demand is a good idea. We might have to add rate-limiting before releasing this patch.

vtnerd:
> We've had a few scanning issues with historical address

The scanner missed transactions? or?

src/util/ownership_test.h
```diff
@@ -0,0 +1,95 @@
...
+    std::optional<db::storage> disk;
+    db::cursor::subaddress_indexes cur;
+    const std::uint32_t max_subaddresses;
+
+    void update_reader();
+  };
+
+  class ownership_test {
+  public:
+    using spend_action = std::function<void(account&, db::spend&&)>;
```
vtnerd: I don't see why these would take a `&&`, theres nothing that can be moved in these types. Guess it doesn't hurt though.

vtnerd:
> The database has been much more stable since the recent corruption fixes, but it's still not 100%. One server died suspiciously close to a block reorg, but this may be a red herring.

This is both good to hear, and frustrating. Another user reported DB issues on 0.3 but never followed up with a detailed log to help narrow down where it occurred.

src/rest_server.cpp
```diff
@@ -1579,6 +1647,24 @@ namespace lws
...
+        cryptonote::blobdata tx_blob;
+        if (!epee::string_tools::parse_hexstr_to_binbuff(req.tx, tx_blob))
+          return {lws::error::bad_client_tx};
+
+        cryptonote::transaction tx;
+        if (!cryptonote::parse_and_validate_tx_from_blob(tx_blob, tx))
+          return {lws::error::bad_client_tx};
+
+        // also add the transaction to the mempool on successful broadcast
+        auto wrapped_resume =
```
vtnerd: My preferred implementation for this would be to change line 1797 to a `.clone()` (which is an atomic reference-count bump), and then in the `else` (success) block in `async_handler::response` the tx blob would be unpacked to a `cryptonote::transaction` and moved to the cache. This keeps the endpoint slightly faster as it doesn't need to allocate memory for another `std::function`, and the memory is kept to a minimum while waiting for the server to respond.

src/mempool.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  std::shared_ptr<const mempool::pool_table> mempool::get_txs() const
+  {
+    const std::lock_guard<std::mutex> lock{mutex_};
+
+    // return the existing snapshot if we have one
+    if (snapshot_) return snapshot_;
+
+    // otherwise make a fresh snapshot
+    snapshot_ = std::make_shared<const pool_table>(state_);
```
vtnerd: There's `boost::flat_unordered_map` and `boost::container::flat_map` which should be much quicker to copy (since it can be done in one allocation). The performance should be close to a `std::unordered_map`, so I wouldn't expect much drop-off.

src/rpc/client.cpp
```diff
@@ -481,6 +482,118 @@ namespace rpc
...
+      {
+        auto json = net::zmq::receive(daemon.get(), ZMQ_DONTWAIT);
+        if (!json)
+        {
+          if (json == net::zmq::make_error_code(EFSM))
+          {
+            MERROR("FSM again: " << json.error());
+            break;
+          }
+          if (json == net::zmq::make_error_code(EAGAIN) || json == net::zmq::make_error_code(EFSM))
```
vtnerd: This checks for `EFSM` but line 533 already checked for that above and hit a `break`.

src/util/ownership_test.cpp
```diff
@@ -0,0 +1,325 @@
...
+    if (match.is_zero())
+      return {}; // keep subaddress disabled servers quick
+
+    auto upserted = disk->update_lookahead(user.db_address(), height, match, max_subaddresses);
+    if (!upserted)
+      return upserted.error();
+    if (0 < *upserted)
+      update_reader(); // update reader after upsert added new addresses
+    else if (*upserted < 0)
+      upserted = {error::max_subaddresses};
```
vtnerd: The `max_subaddresses` error is no longer returned to the caller.

src/mempool.cpp
```diff
@@ -0,0 +1,215 @@
...
+          auto shared_tx = std::make_shared<const cryptonote::transaction>(std::move(tx.tx));
+          txs.insert({tx.tx_hash, std::move(shared_tx)});
+        }
+        pool->reset_txs(std::move(txs));
+        return {};
+      },
+
+      [&client](rpc::minimal_chain_pub&& msg) -> expect<void>
+      {
+        cryptonote::rpc::GetTransactionPool::Request req{};
```
vtnerd: One issue with this code is that it doesn't wait for the previous request to complete, so the `EFSM` code could trigger. It looks like you handle that by ignoring the error, which should work, assuming the server is guaranteed to respond _eventually_.

My understanding of ZMQ: you cannot guarantee the server will respond. The ZMQ framework could send it to `monerod`, which then immediately crashes, and then `monerod` gets restarted. The state machine on the client side is waiting for a response. There's no clearing of state for a re-connection (afaik).

I don't have any recommendations to fix this, as the spillover may be quite large to fix it. The way I handle this in the `rest_server` was just to destroy the ZMQ client object when it hit this error. I _think_ you can do that in the `pool_update_loop` caller.

vtnerd (changes requested):
This is a fairly clean implementation. My only major gripes (in order of importance):

  * A `EFSM` issue where the client assumes server will respond. See comments below for more.
  * The `scanner` always searches the mempool against all active accounts, even when a recently active account just scanned the whole mempool in `get_address_txs` (wasted CPU cycles).
  * The REST thread is now doing more work than I'd like (scan the mempool). The only way around this is a re-design where the scanner pushes matching receives/spends to an object that `get_address_txs` can read.


For the third bullet-point, have you investigated such a design? Or was scanning on-demand the first attempt?

vtnerd:
This became #251 and has been merged.

PR #223, opened 2025-12-20
Various address assignment optimizations

paullinator:
Implement various address assignment optimizations

block_depth_threading: Assign addresses to threads based on their amount of blocks remaining to be scanned.
split_sync_threads: Assign a percentage of threads to be only used for synced addresses
balance_new_addresses: Put newly added addresses to threads whose current blockscanning height most closely matches the birthday height of the incoming address
More details of each optimization can be found in the MD files in docs/ of the relevant commits

### DIFF
--- a/src/client_main.cpp
+++ b/src/client_main.cpp
@@ -246,7 +246,7 @@ namespace
 
         if (!users.empty())
         {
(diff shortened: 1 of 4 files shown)


vtnerd:
@paullinator have you used these in testing? I'm mainly curious about re-balancing some of this after threads catch-up. There's probably enough "re-starts" (import requests, chain reorgs) where it effectively rebalances anyway ... ?

PR #237, opened 2026-02-26
Implemented websocket /feed on HTTP (REST) API

vtnerd:
This is marked as a draft until the client implementation is done. I want to make sure the protocol works before even merging this code.

This design adds a `/feed` endpoint that is _not_ REST, instead the endpoint is a web socket push/stream protocol. The user logs in via address + viewkey, then automatically gets the entire state of transactions from the server. The schema differs from the existing REST api in that the minimum amount of data to support the `wallet_api.h` is sent (the existing REST API had lots of data redundancy).

After that initial sync, the `/feed` interface sends "raw" scanner information to the frontend, via ZMQ internally, including mempool matches and all newly scanned blocks. Reorgs and monerod disconnects are also reported to the clients. If the `/feed` threads are overloaded and miss internal ZMQ messages, this is detected and the stream errors out.

The design supports both JSON and msgpack, the latter being ~10% faster in prior tests.

There are some basic tests, but more would be useful. At this stage I'd rather get a _real_ client so that I can test viability of the design. There will also be a markdown explanation of the final protocol.

### DIFF
--- a/docs/api/feed.md
+++ b/docs/api/feed.md
@@ -1,0 +1,258 @@
+# '/feed' API
+
+The `/feed` websocket API allows for faster synchronization on initalization (msgpack
+with less redundant data compared to [wallet api](/api/wallet/)), and "push"
+updates instead of polling. New block arrivals, new received/spent transactions,
+and reorgs are all pushed to the client in real-time. This should provide an
+enterprise-like experience where the (usually mobile) client is seamlessly
+synchronized with `monerod` (but only if `monero-lws-daemon` is using ZMQ
+PUB/SUB from `monerod`).
+
+## Basics
+The client initiates a `/feed` websocket connection, then the LWS server sends
+the entire tx history, and then finally sends a continous stream of updates from
+`monerod` (ZMQ PUB mempool+blocks messages) to the client.
+
+## Flow
+### Handshake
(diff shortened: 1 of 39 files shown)


vtnerd:
I'm going to "rebrand" this as `/feed` to avoid confusion with web push type stuff. The  REST endpoint for establishing a web push will likely be `/push` instead. 

*if ever implemented, as its more complex than `/feed`

vtnerd:
Force pushed some tweaks to the protocol after implementing the lwsf (frontend) portion. These changes reduce the fields to the bare minimum, and mostly appear working.

A separate issue for "null" tx-hashes was discovered and will be patched separately.

vtnerd:
Mostly renames with a few bug fixes. This matches what the client is now doing.

vtnerd:
Rebased and added docs with schemas for /feed

vtnerd:
A critical fix for stability, caught with ASAN.

vtnerd:
Tweaked the /feed schema. The `id` field of `spend`s are no longer required. This is needed for FCMP++ legacy and carrot view-incoming accounts as the backend can only infer a spend through a CHANGE tx (the output being spent cannot be known).

PR #248, opened 2026-04-08
Add /daemon_status to docs

Keeqler:


docs/api/wallet.yaml
```diff
@@ -18,6 +18,47 @@ tags:
...
+                $ref: '#/components/schemas/daemon_status_response'
+        '405':
+          description: HTTP method is not GET or POST.
+        '500':
+          description: General errors.
+        '503':
+          description: ZMQ/daemon unavailable, timeout, or overload.
+    post:
+      tags:
+        - transaction
```
vtnerd, line 44: I don't think this tag is accurate - it's not related to specific transactions. I would just omit the tags like `/get_version`, or create a new one for both.

docs/api/wallet.yaml
```diff
@@ -18,6 +18,47 @@ tags:
   - name: subaddress
   - name: transaction
 paths:
+  /daemon_status:
+    get:
+      tags:
+        - transaction
```
vtnerd, line 24: Same thing here.

vtnerd (changes requested):

PR #255, opened 2026-05-23
Update Dockerfile (expat, unbound, monero)

vtnerd:
Updates Dockerfile to use newest libs and monero code. This should easily pass CI as `develop` doesn't use these files. Once a single CI passes I will merge and push this all to master to update CI.

The 0.3 Dockerfile will need similar updates.


vtnerd:
Forgot to mention that this updates the `external/monero` pin, so the (build-beginner) build could theoretically fail. So obviously waiting for this to pass through.

EDIT: Oh sorry! build-beginner doesn't use the pin (sorry for the noise)

PR #257, opened 2026-05-23
Changes needed to work with 0.18 branch (including C++14 target) [1.0 beta]

vtnerd:
This contains the necessary changes to make the LWS `v1.0_0.18` branch compatible with `v0.18` branch of `monero`. This also downgrades to C++14 to follow the 0.18 branch (the goal being reducing incompatibilities in C++ version being used).

The C++ version downgrade probably isn't necessary (it seems to work with mixed C++14/17), but it seemed just as easy to be on the safe side. There's only one real concession where a `constexpr` was downgraded to a runtime `static` initialization.

None of this will be ported to the `master` branch as its better to keep that C++17 ready, for the inevitable bump to monero 0.19 tree.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -10,12 +10,12 @@ on:
 
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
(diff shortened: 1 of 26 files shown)


vtnerd:
Force pushed push missing file.

vtnerd:
Another force push - the `external/monero` ref was modified from `master` to `v0.18.5` tag. This will be necessary for the Docker build to eventually work when this PR is merged.

PR #258, opened 2026-05-26
Fix msgpack small negative integer output

vtnerd:
During `monero-project/monero` reviews, an error in the msgpack integer output was discovered. Small negative integers were never output as a single byte, instead taking 2 bytes at minimum. This fixes that issue. A tweak to the read may be needed in the future, but for now the tests indicate it is working.

EDIT: Clarification - the disruption should be minimal as the error simply used more bytes than necessary to encode a value. So most msgpack readers should handle it, but possibly output a warning.

### DIFF
--- a/src/wire/msgpack/write.cpp
+++ b/src/wire/msgpack/write.cpp
@@ -112,8 +112,9 @@ namespace wire
   void msgpack_writer::do_integer(const std::intmax_t value)
   {
     assert(value < 0); // constraint checked in header
-    if (0xe0 < value) // 0xe0 needs to be type `int` to work
+    if (-32 <= value)
     {
+      static_assert(msgpack::ftag_signed::tag() == std::uint8_t(-32), "");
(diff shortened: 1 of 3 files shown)


src/wire/msgpack/write.cpp
```diff
@@ -112,8 +112,9 @@ namespace wire
   void msgpack_writer::do_integer(const std::intmax_t value)
   {
     assert(value < 0); // constraint checked in header
-    if (0xe0 < value) // 0xe0 needs to be type `int` to work
+    if (-32 <= value) // 0xe0 needs to be type `int` to work
```
selsta: comment should be removed

vtnerd: Will fix after remaining CIs complete. Then I'll probably immediately merge.

The failed sanitizer test was/is a timeout issue, the CI server likely too long to execute the /feed code.

PR #260, opened 2026-05-27
Fix 'real' output in msgpack [1.0]

vtnerd:
Real (float/double) msgpack output is not used in the 1.0 LWS branch. However, the `::wire` interface includes it because JSON uses it for reading exchange rates (when enabled).

I therefore decided to "fix" the float/double msgpack output/input in the 1.0 branch just in case some feature gets imported which uses it.

This patch will eventually be included in the `master` branch, but more work is needed to get that branch up-to-date with monero mainline.

### DIFF
--- a/src/wire/msgpack/read.cpp
+++ b/src/wire/msgpack/read.cpp
@@ -387,10 +387,16 @@ namespace wire
 
     const auto read_float = [this](auto value)
     {
-      if (remaining_.size() < sizeof(value))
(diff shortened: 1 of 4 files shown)


vtnerd:
Looks like this `/feed` test timeout is too aggressive for GH actions. This will have to be a separate gross fix to allow more time.

vtnerd:
Force pushed a rebase, and changed the implementation such that the min Boost level did not have to be bumped.

PR #263, opened 2026-06-26
fix: pass regtest flag to initial startup sync + regtest re-org aware [develop]

plowsof:
a fresh sync rejects a regtest fakechain ('Attempting rollback past last checkpoint ... wrong --network?') even with --regtest

### DIFF
--- a/src/db/storage.cpp
+++ b/src/db/storage.cpp
@@ -3171,7 +3171,9 @@ namespace db
 
       // collect all .value() errors
       updated out{};
-      if (get_checkpoints().get_max_height() <= last_update)
+      // regtest hast no checkpoints, so a re-org can occur at any height (including below the
+      // mainnet checkpoint height that get_checkpoints() reports)
+      if (lws::config::regtest || get_checkpoints().get_max_height() <= last_update)
       {
         cursor::blocks blocks_cur;
         cursor::pow    pow_cur;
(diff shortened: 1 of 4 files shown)


plowsof:
closing duplicate of #259

src/db/storage.cpp
```diff
@@ -3171,7 +3171,9 @@ namespace db
 
       // collect all .value() errors
       updated out{};
-      if (get_checkpoints().get_max_height() <= last_update)
+      // regtest hast no checkpoints, so a re-org can occur at any height (including below the
+      // mainnet checkpoint height that get_checkpoints() reports)
+      if (lws::config::regtest || get_checkpoints().get_max_height() <= last_update)
```
vtnerd, line 3176: This function will throw `lmdb::error(MDB_NOTFOUND)` or return `error::blockchain_reorg` when in regtest mode. When the former happens, LWS will shutdown. I doubt this is the intended behavior of the `regtest` flag.

The throw/shutdown occurs when the last block in the update is not in the DB. Only blocks after the last checkpoint are guaranteed to be in the DB, unless `--untrusted-daemon` is also used (which forces the DB to store all block hashes).

I'll need more context into how this is being used before recommending a fix.

vtnerd (changes requested):

PR #264, opened 2026-07-01
Fix integer overflow check in subaddress handling

vtnerd:


vtnerd:
Forgot that I need to update `develop` before doing this.

PR #275, opened 2026-07-11
chore: fix typo in build instructions

jpk68:


vtnerd:
The build errors will obviously unrelated - the advanced mode now requires another lib in the dependency list.

PR #276, opened 2026-07-11
Fix advanced build linking

vtnerd:
Lets all watch the CI.


selsta:
https://github.com/monero-project/monero/pull/10899 is needed to fix CI

PR #278, opened 2026-07-15
Fix `MDB_BAD_RSLOT` in `subaddress_reader::update_reader()` causing permanently missed subaddress outputs

### LINKED ISSUE #277: Subaddress outputs silently missed after `MDB_BAD_RSLOT` in scanner
Scanning an account with subaddress lookahead sometimes permanently misses outputs sent to a
subaddress (primary-address outputs in the same blocks are fine), even after repeated full
rescans. Root cause visible in logs:

```
I Thread 7 processed 1001 blocks(s) @ height 1000 against 1 account(s)
E Subadress lookup failure: MDB_BAD_RSLOT: Invalid reuse of reader locktable slot
W Failed to update lookahead for <address>: common_category():2
I Thread 7 processed 713 blocks(s) @ height 1712 against 1 account(s)
```

`MDB_BAD_RSLOT` = an LMDB reader-lock slot got reused across threads. After it fires, the
account's lookahead update silently fails, and the next batch scans with a broken subaddress
lookup table for that account — some subaddress outputs in the batch are found, others aren't.

**Repro:** multi-threaded scanner, account with `lookahead: {maj_i:1, min_i:10}`, ~1700+ blocks
of history (more than one `get_blocks_fast` batch), transactions to subaddress `0/1` mixed with
primary-address transactions. Compare `/get_address_txs` to what was actually sent.

**Env:** `ghcr.io/vtnerd/monero-lws:master` @ `00d948e1a70c5ae6cb7187342982199999d001eb`,
`monero_version_full: 0.18.1.0-be4493d1b`, 8 scan threads.

**Expected:** all owned outputs detected regardless of batch boundaries, or a hard
error/retry instead of silently continuing after `MDB_BAD_RSLOT`.

mahnunchik:
fix https://github.com/vtnerd/monero-lws/issues/277

## Problem

`update_reader()` does:
```cpp
reader = disk.start_read();
```
The right-hand side opens a *new* LMDB read transaction before the assignment destroys the
*old* one held in `reader` — for an instant the same thread holds two live read txns on one
environment, which LMDB's default (non-`MDB_NOTLS`) reader-slot tracking doesn't support and
reports as `MDB_BAD_RSLOT`.

This function only runs when the subaddress lookahead window expands (`update_lookahead()`,
`*upserted > 0`), so the failure hits deterministically: whichever tx is the first to reach a
new minor index trips it. Once it fails, `reader` is left permanently invalid, and
`find_subaddress()` then silently returns `MDB_NOTFOUND` for the rest of that scan pass (the
caller treats `MDB_NOTFOUND` as "not ours", nothing logged) — one glitch permanently blinds the
scanner to further subaddress outputs in that pass, on every fresh rescan of the same chain.

## Fix

Release the old read txn before opening the new one, instead of letting both exist briefly:
```cpp
reader = expect<db::storage_reader>{common_error::kInvalidArgument};
reader = disk.start_read();
```

## Test

`scanner.test.cpp`'s `lws::scanner::run (with lookahead)` section already exercises this exact
path (a tx pays a subaddress at minor index 2, outside the account's initial lookahead window,
forcing a mid-scan `update_reader()`) — but its assertions were calibrated against the buggy
output rather than correct behavior: `outputs.count() == 6` and a frozen
`index_range{0,2}`. With the fix, all 7 expected outputs are recorded (verified field-by-field
against the test's own `expected` map) and the lookahead range correctly expands to `{0,3}`.
Updated both assertions to match. Both the bug and the miscalibrated test were introduced
together in 16111ca (subaddress lookahead support) — the test was written against the
already-buggy implementation, so it never caught the regression.

### DIFF
--- a/src/server_main.cpp
+++ b/src/server_main.cpp
@@ -387,7 +387,7 @@ namespace
       mempool = std::make_shared<lws::mempool>();
     }
 
-    auto client = scanner.sync(ctx.connect().value(), prog.untrusted_daemon).value();
+    auto client = scanner.sync(ctx.connect().value(), prog.untrusted_daemon, prog.regtest).value();
 
     lws::rest_server server{
       epee::to_span(prog.rest_servers),
--- a/tests/unit/scanner.test.cpp
+++ b/tests/unit/scanner.test.cpp
@@ -941,7 +941,7 @@ LWS_CASE("lws::scanner::sync and lws::scanner::run")
 
         auto reader = MONERO_UNWRAP(db.start_read());
         auto outputs = MONERO_UNWRAP(reader.get_outputs(lws::db::account_id(1)));
-        EXPECT(outputs.count() == 6);
+        EXPECT(outputs.count() == 7);
         auto output_it = outputs.make_iterator();
         for (auto output_it = outputs.make_iterator(); !output_it.is_end(); ++output_it)
         {
@@ -1000,7 +1000,7 @@ LWS_CASE("lws::scanner::sync and lws::scanner::run")
 
         {
           const std::vector<lws::db::subaddress_dict> expected_range{
-            {lws::db::major_index(0), {{lws::db::index_range{lws::db::minor_index(0), lws::db::minor_index(2)}}}}
+            {lws::db::major_index(0), {{lws::db::index_range{lws::db::minor_index(0), lws::db::minor_index(3)}}}}
           };
           EXPECT(MONERO_UNWRAP(reader.get_subaddresses(lws::db::account_id(1))) == expected_range);
         }
--- a/src/util/ownership_test.cpp
+++ b/src/util/ownership_test.cpp
@@ -305,6 +305,11 @@ namespace lws
 
   void subaddress_reader::update_reader()
   {
+    // Release the current read txn before opening the next one - `reader = disk.start_read()`
+    // would otherwise begin the new LMDB read txn (inside start_read()) while the old one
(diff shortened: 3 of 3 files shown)


tests/unit/scanner.test.cpp
```diff
@@ -941,7 +941,7 @@ LWS_CASE("lws::scanner::sync and lws::scanner::run")
 
         auto reader = MONERO_UNWRAP(db.start_read());
         auto outputs = MONERO_UNWRAP(reader.get_outputs(lws::db::account_id(1)));
-        EXPECT(outputs.count() == 6);
+        EXPECT(outputs.count() == 7);
```
vtnerd, line 944: Yikes, a serious bug in the tests.

src/server_main.cpp
```diff
@@ -387,7 +387,7 @@ namespace
       mempool = std::make_shared<lws::mempool>();
     }
 
-    auto client = scanner.sync(ctx.connect().value(), prog.untrusted_daemon).value();
```
vtnerd, line 390: This is unrelated, but I'll allow it.

vtnerd (approved):
Next time issue the PR onto `develop`. I stage commits there first as `master` pushes an alpha docker build which some people use for testing.

src/server_main.cpp
mahnunchik: Sorry, this is from previous PR https://github.com/vtnerd/monero-lws/pull/259/changes

PR #281, opened 2026-07-30
Fix shared locking in lmdb map resizing

vtnerd:
Alternative to #280 that starts with a smaller map size, and uses a boost provided "fair" `shared_lock` (as opposed to the gross spin lock).


mahnunchik:
@vtnerd I confirm that this fix solves the issue https://github.com/vtnerd/monero-lws/pull/280

With default settings, the LWS node was able to scan 100 almost simultaneous scan requests without any problems.

mahnunchik:
I still cannot find the real cause. Now, at random times, hundreds of messages appear in the logs:
```
Unable to lookup webhook on tx in pool: MDB_BAD_RSLOT: Invalid reuse of reader locktable slot
```
But scanning continues without hanging

vtnerd:
This is almost certainly a separate issue. There was a recent PR related to this, but it seems like the issue persists.

vtnerd:
Yes, this is a bug with subaddresses+webhooks. Thanks for catching this. It'll have to be re-worked somehow.

vtnerd:
Ill merge this and start a new PR for this new issue.

PR #283, opened 2026-08-07
Fix mempool webhook read during subaddress scanning

vtnerd:
@mahnunchik reported in #281 that a LMDB read error can occur when trying to lookup webhooks for mempool txes. I was able to reproduce the issue, and write a unit test that verifies the fix. So we get a bug fix + a 0-conf webhook test!

Watching the CI for this, as this has only been tested on Ubuntu 22.04.


vtnerd:
Guess I'm bumping the monero pin too, as this is difficult to change.

PR #284, opened 2026-08-12
Update libexpat and boost in Docker builds

vtnerd:
The CI is useless for this, as it only runs on master and release branches. But posting this anyway for visibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,11 +41,11 @@ ENV USE_SINGLE_BUILDDIR 1
 ENV BOOST_DEBUG         1
 
 # Build expat, a dependency for libunbound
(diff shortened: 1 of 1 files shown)


vtnerd:
Force pushed an update to the monero external "pin" which is needed for this Dockerfile. The master branch already has this.

Not sure why _some_ of the CI's failed last time.

nahuhh:
Ci failing look related to monero master now pulling in fcmp. You might want to switch the submodule to release-v0.19, or add rust to the deps

vtnerd:
Yup, fixed this in #285

vtnerd:
We probably need a better error message too, as it occurs during building with an obscure message.

PR #286, opened 2026-08-17
Bypass checkpoint gate for regtest chain reorgs

everoddandeven:


vtnerd (commented):
I don't fully understand why you need to test reorgs past a checkpoint. You can easily test 100+ reorgs without `regtest` ... so what are you testing?

everoddandeven:
> I don't fully understand why you need to test reorgs past a checkpoint. You can easily test 100+ reorgs without `regtest` ... so what are you testing?

I'm testing the light wallet implementation I made on [monero-cpp](https://github.com/woodser/monero-cpp/pull/73), and through python bindings tests I noticed stale indices/transactions in wallet info provided by lws after a block reorg on a regtest environment (checkpoints have no meaning there).

PR #289, opened 2026-09-05
src: fix a few broken checks

jpk68:


src/wire/read.h
```diff
@@ -307,8 +307,8 @@ namespace wire_read
     if (max_element_count < count)
       throw_exception(wire::error::schema::array_max_element, "", nullptr);
 
-    // also checked by derived formats when count is known
-    if (min_element_size && (source.remaining().size() / min_element_size) < count)
+    const std::size_t effective_min_element_size = min_element_size ? min_element_size : 1;
```
vtnerd: I prefer `std::max(min_element_size, size_t(1))`.

This is intentional (not a mistake or oversight); this would break with ASN.1 PER (or similar crazy encoding schemes). Otoh its unlikely that LWS ever supports a format where an element can take up zero bytes.

src/db/storage.cpp
```diff
@@ -2371,7 +2371,7 @@ namespace db
               const auto this_minor = add_and_clamp(minor - 1, to_uint(receipient->min_i));
 
               // Quick Sanity Check before vector expansion
-              if (minor && std::numeric_limits<std::uint32_t>::max() < needed / minor)
+              if (std::numeric_limits<std::uint32_t>::max() / minor < needed)
```
vtnerd, line 2374: Thought I got all of these, guess not.

src/rpc/scanner/read_commands.h
```diff
@@ -128,6 +128,12 @@ namespace lws { namespace rpc { namespace scanner
 
           std::memcpy(std::addressof(self_->next_), self_->read_buf_.data(), sizeof(self_->next_));
           static_assert(std::numeric_limits<header::length_type::value_type>::max() <= std::numeric_limits<std::size_t>::max());
+          if (max_header_length < self_->next_.length.value())
```
vtnerd: I'd have to step through this again, but this is likely going to be a nack for this change.

The problem is that there's a command where every account sends its updates. If this command exceeds `max_header_length`, then the remote scan is effectively blocked.

This feature (for better or worse) requires trusted endpoints. You cannot send your scan to some remote process, and _not_ trust it.

This does guard against unintentionally large commands, so its not without merit.

src/wire/msgpack/read.cpp
```diff
@@ -441,7 +441,8 @@ namespace wire
       read_count<msgpack::ftag_array, msgpack::array_types>(error::schema::array);
     if (limits<std::size_t>::max() - tags_remaining_ < upcoming)
       WIRE_DLOG_THROW_(error::msgpack::max_tree_size);
-    if (min_element_size && (remaining_.size() / min_element_size) < upcoming)
+    const std::size_t effective_min_element_size = min_element_size ? min_element_size : 1;
```
vtnerd: Another `std::max(min_element_size, size_t(1))` syntax.

And this one is an improvement, as msgpack never has a 0 byte element.

vtnerd (changes requested):

vtnerd (approved):

